PR #25, opened 2016-02-26
Add display for request/response headers fixes #16

### LINKED ISSUE #16: [HttpServer] Display request/response headers
Update UI to render Request/Response headers in line with current Execution tab:
 - [x] Update execution tab to show headers
 - [x] Get feedback/review of final UI design
 - [ ] Add unit tests**

** Note we currently don't have unit tests for the UI. This point will need to reviewed depending on the state of the client unit tests.

**Design:**
It is expected that the Request Headers will be displayed at the top of the Execution Tab and Response Headers at the bottom. The section should have both a summary view and details view (as we need to be careful how much vertical space we take up). The keys of the headers should be highlighted/bolded, and the values parsed/trimmed/formatted as it makes sense.

avanderhoorn:
Initial cut of adding the request/response headers to the execution screen.

screen shot 2016-02-25 at 6 29 06 pm

Now awaiting feedback on what the design is - @nikmd23

### DIFF
--- a/fake/fake-extension.js
+++ b/fake/fake-extension.js
@@ -436,6 +436,17 @@ var generateMvcRequest = (function() {
             payload.requestPath = source.path;
             payload.requestQueryString = source.queryString;
             payload.requestUrl = 'http://localhost:5000' + source.path + defaultOrEmpty(source.queryString);
+            payload.requestHeaders = {
+                'Server': 'GitHub.com',
+                'Date': 'Mon, 02 Nov 2015 06:39:26 GMT',
+                'Content-Type': 'text/html; charset=utf-8',
+                'Transfer-Encoding': 'chunked',
+                'Status': '200 OK',
+                'Cache-Control': 'no-cache',
+                'Strict-Transport-Security': 'max-age=31536000; includeSubdomains; preload',
+                'Vary': 'Accept-Encoding',
+                'Content-Encoding': 'gzip'
+            }; 
             
             MessageGenerator.support.beforeTimings('request', payload, source.dateTime);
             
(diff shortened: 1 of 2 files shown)


avanderhoorn:
Thinking out loud....

I'm wondering if we either hide the headers within a expandable section and show a summary like what we do in v1.
-or-
Riffing on the expander idea we just show the header keys and show the value when the user opens it 
-or-
We could have a hybrid of what we have at the moment and an expander where the white list of headers that we want to always show and the rest are in the expander.

nikmd23:
I propose we move the headers to a "Request" tab. The layout you have now is good, it just doesn't have much to do with "execution".

There will be more things added to the request tab as we go, but it's not particularly interesting ATM. Thoughts?

avanderhoorn:
I'm open to that... any ideas on what type of data we would see in a future Request tab besides headers? Thinking out loud, would we put the resources that this response triggered to load, etc?

nikmd23:
Yes, that's a good example.

Also I'm thinking things like:
- Compression ratios
- Protocol used
- Request bodies
- Cookies
- CSP Violations
- Data from [HttpLint](http://zamez.org/httplint), [WhatDoesMySiteCost](https://whatdoesmysitecost.com/) and/or [RedBot](https://redbot.org/)
- Etc

Way down the line, I could even see some simple "edit and replay request" functionality on this page.

avanderhoorn:
Great!

avanderhoorn:
This means that I will close this PR and open up a new one. Do you have any thoughts on what the design could/should look like to support this vision?

PR #29, opened 2016-02-29
changes to allow mongodb messages to show up in the messages tab

mike-kaufman:
Necessary changes to get the mongoDB messages to show up in the messages tab.


mike-kaufman:
spoke with Anthony and this is the wrong fix.  Correct fix is a bug in Glimpse.server.

PR #30, opened 2016-03-01
[Http] Add new request tab to start showing http data

avanderhoorn:
Initial cut of adding the request/response headers to the execution screen.


nikmd23:
One small nit-pick, Change the label at the top from "Http Request/Response" to "Web Request/Response".

Other than that: :shipit:

avanderhoorn:
If any further changes are needed they can be picked up in another PR.

PR #31, opened 2016-03-01
Adding basic UI for mongo messages

mike-kaufman:
Adding very basic UI for mongo messages.  There's still much to do here, but I wanted to get this basic very basic stuff checked in so we have something to display. There's also lots of opportunity to refactor how the display logic.  That's left as a future exercise.

### DIFF
--- a/src/request/components/request-detail-panel-execution.jsx
+++ b/src/request/components/request-detail-panel-execution.jsx
@@ -50,30 +50,6 @@ var getMessages = (function() {
     }
 })();
 
-var RequestHeaders = React.createClass({
-    render: function() {
-        return (
-            <div>
-                <div className="tab-section">{this.props.title}</div>
-                <div className="tab-section tab-section-execution-headers">
-                    <div className="flex flex-row flex-inherit tab-section-header">
-                        <div className="tab-title col-2">Variable</div>
-                        <div className="tab-title col-8">Value</div>
-                    </div>
-                    <div className="tab-section-boxing tab-section-listing">
-                        {_.map(this.props.headers, function(value, key) {
-                            return (<section className="flex flex-row">
-                                    <div className="col-2">{key}</div>
-                                    <div className="col-8">{value}</div>
-                                </section>);
-                        })}
-                    </div>
-                </div>
-            </div>
-        );
-    }
-});
-
 var MongoCommandMixin = {
     getInitialState: function() {
         return { show: false };
@@ -259,7 +235,7 @@ var CommandList = React.createClass({
             // process action
             if (commandItems.length > 0) {
                 content = (
-                        <div className="tab-section tab-section-execution-command">
+                        <div className="tab-section tab-section-boxed tab-section-execution-command">
                             <div className="flex flex-row flex-inherit tab-section-header">
                                 <div className="tab-title col-9">SQL Query</div>
                             </div>
@@ -288,7 +264,7 @@ var CommandList = React.createClass({
          // process action
          if (mongoOperations.length > 0) {
                 content = (
-                        <div className="tab-section tab-section-execution-command">
+                        <div className="tab-section tab-section-boxed tab-section-execution-command">
                             <div className="flex flex-row flex-inherit tab-section-header">
                                 <div className="tab-title col-9">MongoDB</div>
                             </div>
@@ -422,7 +398,7 @@ module.exports = React.createClass({
                 var routePath = beginRequestPayload ? (<span><span>{beginRequestPayload.requestPath}</span><span>{beginRequestPayload.requestQueryString}</span></span>) : '';
             
                 route = (
-                        <div className="tab-section tab-section-execution-route">
+                        <div className="tab-section tab-section-boxed tab-section-execution-route">
                             <div className="flex flex-row flex-inherit tab-section-header">
                                 <div className="tab-title col-9">Route</div>
                             </div>
@@ -444,7 +420,7 @@ module.exports = React.createClass({
                 }
                 
                 action = (
-                        <div className="tab-section tab-section-execution-action">
+                        <div className="tab-section tab-section-boxed tab-section-execution-action">
                             <div className="flex flex-row flex-inherit tab-section-header">
                                 <div className="tab-title col-9">Action</div>
                             </div>
@@ -475,7 +451,7 @@ module.exports = React.createClass({
                     var componentCommands = <CommandList beforeExecuteCommandMessages={beforeExecuteCommandMessages} afterExecuteCommandMessages={afterExecuteCommandMessages} mongoDBMessages={mongoDBMessages} beginMessage={beforeViewComponetMessage} endMessage={afterViewComponetMessage} />;
                     
                     return (
-                            <div className="tab-section tab-section-execution-component">
+                            <div className="tab-section tab-section-boxed tab-section-execution-component">
                                 <div className="flex flex-row flex-inherit tab-section-header">
                                     <div className="tab-title col-9">View Component</div>
                                 </div>
@@ -500,7 +476,7 @@ module.exports = React.createClass({
                 }
             
                 view = (
-                        <div className="tab-section tab-section-execution-view">
+                        <div className="tab-section tab-section-boxed tab-section-execution-view">
                             <div className="flex flex-row flex-inherit tab-section-header">
                                 <div className="tab-title col-9">View Result</div>
                             </div>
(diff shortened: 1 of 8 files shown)


src/request/components/request-detail-panel-execution.jsx
```diff
@@ -145,11 +272,59 @@ var CommandList = React.createClass({
...
+       item = <MongoReadCommand key={mongoDBMessage.id} mongoOperation={mongoDBMessage} />
+    }
+    else if (_.intersection(mongoDBMessage.types, ['data-mongodb-update']).length === 1){
+       item = <MongoUpdateCommand key={mongoDBMessage.id} mongoOperation={mongoDBMessage} />  
+    }
+    else if (_.intersection(mongoDBMessage.types, ['data-mongodb-delete']).length === 1){
+       item = <MongoDeleteCommand key={mongoDBMessage.id} mongoOperation={mongoDBMessage} />
+    }
+    return item;
+}
```
avanderhoorn: - `intersection` - We have some helpers which might make this logic a bit easier... https://github.com/Glimpse/Glimpse.Client.Prototype/blob/master/src/request/util/request-message-processor.js#L25-L56. It also turns out that we have the data sliced and diced in a few different ways, so hopefully that might simplify some of the logic here.
 - Next iteration - from here we should be able to do the above based off a list of registered components that are bound to a given message type. This means that we slide SQL Commands, etc into the mix more easily.

src/request/components/request-detail-panel-execution.jsx
```diff
@@ -70,6 +74,128 @@ var RequestHeaders = React.createClass({
...
+                    <div className={containerClass} onClick={this.onClick}>
+                        {this.getDetails()}
+                        <div className="tab-execution-hidden-gradient"></div>                                                
+                    </div>
+                </div>
+            );
+    
+        return content;
+     }
+};
```
avanderhoorn: Nice use of a `mixin` for this work.

webpack.config.js
```diff
@@ -69,7 +69,7 @@ module.exports = {
     plugins: [
         new webpack.DefinePlugin({
             DIAGNOSTICS: true,
-            FAKE_SERVER: true
+            FAKE_SERVER: false
```
avanderhoorn: This should still be `true`. The default for the client is to use the fake server. That way any dev being done here can occur in relative isolation from the server.

src/request/components/request-detail-panel-execution.jsx
```diff
@@ -329,7 +518,7 @@ module.exports = React.createClass({
             //process post action commands
             var postCommands = null;
             if (afterActionResultMessage && endRequestMessage) { 
-                postCommands = <CommandList beforeExecuteCommandMessages={beforeExecuteCommandMessages} afterExecuteCommandMessages={afterExecuteCommandMessages} beginMessage={afterActionResultMessage} endMessage={endRequestMessage} isRoot={true} />            
+                postCommands = <CommandList beforeExecuteCommandMessages={beforeExecuteCommandMessages} afterExecuteCommandMessages={afterExecuteCommandMessages} mongoDBMessages={mongoDBMessages} beginMessage={afterActionResultMessage} endMessage={endRequestMessage} isRoot={true} />
```
avanderhoorn: Ideally, we could just pass all `data` messages in here, but this is an initial cut so its fine for now.

src/request/components/request-detail-panel-execution.jsx
```diff
@@ -216,15 +392,28 @@ module.exports = React.createClass({
...
+            var mongoDBMessages = [];
+            var m = [message.dataMongodbInsert, message.dataMongodbRead, message.dataMongodbUpdate, message.dataMongodbDelete];
+            for ( var i = 0; i<m.length;i++) {
+                if ( m[i] && m[i].length > 0) {
+                    mongoDBMessages = mongoDBMessages.concat(m[i]);
+                }
+            }
+            if ( mongoDBMessages) {
+                mongoDBMessages = mongoDBMessages.sort(function(a,b) { return a.ordinal - b.ordinal;});
+            }
```
avanderhoorn: Can the above be put into a function? Just cleans things up a little... even though its changing later on. It helps isolate the logic and whats going on as future changes come in.

src/request/components/request-detail-panel-execution.jsx
```diff
@@ -70,6 +74,128 @@ var RequestHeaders = React.createClass({
...
+        var mongoOperation = this.props.mongoOperation;
+        var details =
+        <div> 
+            <div>Operation: {mongoOperation.payload.operation}</div>
+            <div>Mongo DB URI: {mongoOperation.payload.connectionHost}:{mongoOperation.payload.connectionPort}/{mongoOperation.payload.database}/{mongoOperation.payload.collection} </div>
+            <div>matched count: {mongoOperation.payload.matchedCount}</div>
+            <div>modified count: {mongoOperation.payload.modifiedCount}</div>
+            <div>upserted count: {mongoOperation.payload.upsertedCount}</div>
+            <div>query: {JSON.stringify(mongoOperation.payload.query)}</div>
+            <div>options: {JSON.stringify(mongoOperation.payload.options)}</div>
```
avanderhoorn: Any reason for some being uppercase and some being lower?

mike-kaufman:
screenshot


Here are two screenshots.  The top one is showing data for a read.  The bottom for an insert.  The bottom insert message is showing up twice, and that's a bug.

src/request/components/request-detail-panel-execution.jsx
```diff
@@ -70,6 +74,128 @@ var RequestHeaders = React.createClass({
...
+    
+    getDetails: function() {
+        var mongoOperation = this.props.mongoOperation;
+        var details =
+        <div> 
+            <div>Operation: {mongoOperation.payload.operation}</div>
+            <div>Mongo DB URI: {mongoOperation.payload.connectionHost}:{mongoOperation.payload.connectionPort}/{mongoOperation.payload.database}/{mongoOperation.payload.collection} </div>
+            <div>query: {JSON.stringify(mongoOperation.payload.query)}</div>
+            <div>options: {JSON.stringify(mongoOperation.payload.options)}</div>
+        </div>;
```
avanderhoorn: If the way we are displaying this data is as a simple list, when we switch over to "data command" strategy (where we have templates bound to message types), we could have a view model defined which has the `title` and the `value`. That way we could just iterate through. For more complex cases, we could have an actual component. Again this can be done in the future.

avanderhoorn:
Great initial cut!!! I'm assuming the above screen shots are showing the details in their `open` state. Any chance we can get a closed one capture as well?

avanderhoorn:
Also, if you can get a copy of a mongo message, lets get it into the data generator so that we can iterate on the design a little faster.

src/request/components/request-detail-panel-execution.jsx
mike-kaufman: me being sloppy.  I'll fix.

src/request/components/request-detail-panel-execution.jsx
mike-kaufman: Agreed.  Was thinking along the same lines.

src/request/components/request-detail-panel-execution.jsx
mike-kaufman: fixed.

avanderhoorn:
I've updated the messages being generated and as you will see when you look at any of the `/` or `/ShoppingCart/` pages, you will see the messages showing up. In the case of `/ShoppingCart/, there are currently some layout issues which seem a little weird. I'm going to take a look at it next as to whats going on there.

avanderhoorn:
Ok I've made some changes to fix a few bugs that where in there and get things up and running with RAD. There is one minor problem with RAD that occurs in cases where queries happen outside of an action and we get a `NaN` but that isn't a problem with this PR. Also, its worth noting that this PR has a limitation where data between SQL and Mongo queries isn't shown in the order of execution. This will need to be addressed in a future PR where we deal with commands a bit better in general.

nikmd23:
This is looking pretty good! It's certainly meeting my expectations for this very early cut.

src/request/components/request-detail-panel-execution.jsx
mike-kaufman: We'll address this with a follow-on set of changes.

src/request/components/request-detail-panel-execution.jsx
mike-kaufman: Yeah, we'll address this with a follow-on set of changes.

mike-kaufman:
y

This is an insert in its closed form.

PR #39, opened 2016-03-09
[HttpServer] Fixes #28 - React to breaking Web schema changes

### LINKED ISSUE #28: [HttpServer] React to breaking Web schema changes
As per [this](https://github.com/Glimpse/Messages/pull/8/files) update we have some breaking schema changes coming down. This means changes to at least the following sections of the app:
 - [ ] Update RAD to publish messages that match the update
 - [ ] Update message [converters](https://github.com/Glimpse/Glimpse.Client.Prototype/blob/master/src/request/repository/converter/request-converter.js#L140-L159) to adjust how indexes are pulled out 
 - [ ] Update the UI (summary and details) to pickup the new names

avanderhoorn:
This brings the client inline with the latest HttpServer schema changes. Note, this change also effects HUD. Fixes #28.


fake/fake-extension.js
```diff
@@ -830,17 +825,20 @@ var generateMvcRequest = (function() {
                     
                     var payload = message.payload;
                     if (type == 'begin-request') {
```
philliphoff: I believe the new message types are `web-request` and `web-response`.

fake/fake-extension.js
```diff
@@ -830,17 +825,20 @@ var generateMvcRequest = (function() {
...
                     if (type == 'end-request') {
-                        request._responseStatusCode = payload.responseStatusCode; 
+                        request._responseStatusCode = payload.statusCode; 
                         request._responseStatusText = source.statusText;
                         request._responseContentCategory = source.contentCategory;  
                         
-                        payload.responseStatusText = request._responseStatusText;  
-                        payload.responseContentCategory = request._responseContentCategory;
+                        payload.statusText = request._responseStatusText;  
+                        payload.contentCategory = request._responseContentCategory;
```
philliphoff: The `statusText` and `contentCategory` aren't parts of the message payload spec.  Is this for another purpose?

fake/fake-extension.js
```diff
@@ -830,17 +825,20 @@ var generateMvcRequest = (function() {
                     
                     var payload = message.payload;
                     if (type == 'begin-request') {
-                        request._requestStartTime = payload.requestStartTime;
-                        request._requestMethod = payload.requestMethod;
-                        request._requestUrl = (payload.requestPath || '') + (payload.requestQueryString || '');
+                        payload.path = source.path;
+                        payload.query = source.query;
+
```
philliphoff: The request message no longer has a `query` payload, in favor of the `url` property.

fake/fake-extension.js
avanderhoorn: Correct. This is needed for internal purposes. The client adds additional derived data.

fake/fake-extension.js
avanderhoorn: Correct. This is needed for internal purposes. The client adds additional derived data.

fake/fake-extension.js
avanderhoorn: Good catch. Will update.

src/request/repository/converter/request-converter.js
```diff
@@ -139,19 +140,25 @@ var getStatusCodeText = (function() {
...
+        request._requestMethod = payload.method;
+        request._requestUrl = payload.path + payload.query;
         request._requestIsAjax = payload.requestIsAjax;
     }
     if (type == 'end-request') {
-        request._responseStatusCode = payload.responseStatusCode; 
-        request._responseStatusText = getStatusCodeText(payload.responseStatusCode);
-        request._responseContentType = payload.responseContentType;
-        request._responseContentCategory = getContentTypeCategory(payload.responseContentType);  
+        var contentType = payload.headers && payload.headers['Content-Type'];
```
philliphoff: Header names are case-insensitive; you might need to also check for `'content-type'`.

src/request/repository/converter/request-converter.js
```diff
@@ -139,19 +140,25 @@ var getStatusCodeText = (function() {
...
     if (type == 'begin-request') {
-        request._requestStartTime = payload.requestStartTime;
-        request._requestMethod = payload.requestMethod;
-        request._requestUrl = (payload.requestPath || '') + (payload.requestQueryString || '');
+        var url = parse(payload.url);
+        payload.path = url.pathname;
+        payload.query = url.query;
+        
+        request._requestStartTime = payload.startTime;
+        request._requestMethod = payload.method;
```
philliphoff: The message spec doesn't include a `method` property; that may be an oversight as it can't be derived from the URL.  We might want to talk to Nik about adding it.

philliphoff: Oh, wait, nevermind.  I was looking at the wrong message.  The request does include the HTTP method.

src/request/repository/converter/request-converter.js
avanderhoorn: You bring up an interesting point. Slight pain here... in theory there are a bunch of different ways you could capitalize the titles/properties. Unfortunately there isn't any nice way that I know of to access/index a property in a case insensitive way. We could loop over each property in the object and then test a case insensitive match, but that feels very ugly. This might mean its worth promoting `ContentType` into its own first level property. @nikmd23

mike-kaufman: isn't the right thing here to loop over all of the payload properties & normalize them to lower-case strings?

mike-kaufman:
other than case-sensitive header names, this looks good to me.

src/request/repository/converter/request-converter.js
nikmd23: I'm not necessarily opposed to promoting `Content-Type` to avoid this issue, but it's a bit of a slippery slope. `Content-Length` could be next, followed by who knows what, as we keep bumping into use cases where we want to do more than just display a value.

All that said, I agree with @mike-kaufman here. Let's just solve the problem once and for all by normalizing the keys. (But don't throw away the original casing since we'll want to display it that way!)

Apparently [lodash makes this easy to do](https://lodash.com/docs#transform) (is that already a dependency?): 

```javascript
var normalizedHeaders = _.transform(headers, function (result, value, key) {
    result[key.toLowerCase()] = value;
});
```

avanderhoorn: For the time being, I'm going to get this working by just pulling up this one property in a case insensitive way. Once we get more use cases where we need multiple header values to be processed we can reevaluate - as copying the object has other implications when it comes to local storage, etc.

mike-kaufman: what do you have in mind by other implications?  just memory usage? memory usage is linear with the number of headers, so it shouldn't be too bad, right?  Or are there other concerns you had in mind?  If lower-casing is as simple as that code that @nikmd23 cites above, I think we should just do that, as it solves this issue for all of our headers.  

lmk if I'm missing something

avanderhoorn:
Ok I've made the last of the changes/updates. Any other feedback?

src/request/repository/converter/request-converter.js
avanderhoorn: If I create a copy of the whole headers object, one concern is memory usage, the other being disk space. Currently we keep a copy of the last x messages the client receives in local storage. Because of this fact, we have been very careful about circular references and the data duplication/normalization. 

An example, messages are indexed via their `id`, as well as by message `type`s. The former "stores" the actual actual message and the later just as an array of `id`s by message `type. This means that if you want to get all the message objects for a given type, you need to use the array of `id`s to then lookup the messages. Even though this is a bit of a pain (which is mitigated with helper functions), it means that I can serialize the whole request object without any worries.

In this cases with headers, if we create a copy of all the headers and store that in the message or against the request, not only are the key values duplicated but the data as well. This could be mitigated in various ways, but given that the current implementation isn't really tailored towards storing too much normalized data, if we don't have any other part of the system which needs to do any programatic work on any other header, it would be adding in complexity where we don't need it atm. 

Does this make sense?

mike-kaufman: ok, makes sense.

mike-kaufman:
OK, lgtm.

PR #40, opened 2016-03-14
[Middleware] Fixes #19 - Update UI to show basic activity

### LINKED ISSUE #19: [Middleware] Update UI to show basic activity
Update UI to render basic middleware data in line with current Execution tab:
 - [ ] Update execution tab to show basic Middleware data (i.e. name, order, duration, etc)
 - [ ] Get feedback/review of final UI design
 - [ ] Add unit tests**

** Note we currently don't have unit tests for the UI. This point will need to reviewed depending on the state of the client unit tests.

**Design:**
Currently the design has the header middleware showing up under the request headers and above the current Connect/MVC data, and the footer middleware showing up above the response headers and below the current Connect/MVC data. We expect for the data to be nested, to help enforce the ordering of middleware.

philliphoff:
This is an initial stab at showing middleware as a separate item within the Execution tab.  The middleware also nests commands (e.g. Mongo DB calls) executed within it. Fixes #19.

screen shot 2016-03-14 at 12 05 22 pm


philliphoff:
Associated with [this middleware task](https://github.com/Glimpse/Glimpse.Node.Prototype/issues/89).

src/request/components/request-detail-panel-execution.jsx
```diff
@@ -336,6 +338,60 @@ var ParamaterList = React.createClass({
...
+                switch (result) {
+                    case 'next': result = 'Next'; break;
+                    case 'end': result = 'End'; break;
+                    case 'error': result = 'Error'; break;
+                }
+                
+                return (
+                        <div className="tab-section-boxing">
+                            <section className="flex flex-row flex-inherit flex-base tab-section-item">
+                                <div className="tab-execution-important col-8">{middlewareEndPayload.name} &nbsp; <span className="text-minor">({result})</span></div>
```
avanderhoorn: Lets remove `tab-execution-important` from here. The action/callback title and view names are more important in this context.

src/request/components/request-detail-panel-execution.jsx
```diff
@@ -336,6 +338,60 @@ var ParamaterList = React.createClass({
...
+                var middlewareEndMessage = middlewareEndMessages[middlewareStartMessage.payload.correlationId];
+                var middlewareEndPayload = middlewareEndMessage.payload;
+                
+                var result = middlewareEndPayload.result;
+                
+                switch (result) {
+                    case 'next': result = 'Next'; break;
+                    case 'end': result = 'End'; break;
+                    case 'error': result = 'Error'; break;
+                }
```
avanderhoorn: Why don't we just [title case](https://lodash.com/docs#startCase) this? It would work for the above cases as well as any other currently fall through cases.

src/request/components/request-detail-panel-execution.jsx
```diff
@@ -397,6 +455,13 @@ module.exports = React.createClass({
                 preCommands = <CommandList beforeExecuteCommandMessages={beforeExecuteCommandMessages} afterExecuteCommandMessages={afterExecuteCommandMessages} mongoDBMessages={mongoDBMessages} beginMessage={beginRequestMessage} endMessage={beforeActionInvokedMessage} isRoot={true} />            
             }
             
+            // process middleware
+            var middleware = null;
+
```
avanderhoorn: nit, we should be able to remove this.

fake/fake-extension.js
```diff
@@ -170,6 +170,13 @@ var seedMvcActions = (function() {
...
+                            { access: 'middleware-start', correlationId: '234567891', name: 'middleware2', startTime: '2016-02-29T14:54:40.073 -0800' },
+                            { access: 'mongo', type: 'data-mongodb-read', operation: 'toArray', query: {}, startTime: '2016-02-29T14:54:40.073 -0800', options: { find: 'MusicStore.AlbumCollection', limit: 0, skip: 0, query: {}, slaveOk: true, readPreference: { mode: 'primary' } }, connectionHost: 'localhost', connectionPort: 27017, database: 'MusicStore', collection: 'AlbumCollection' },
+                            { access: 'middleware-end', correlationId: '234567891', name: 'middleware2', endTime: '2016-02-29T14:54:41.073 -0800', duration: 1000, result: 'next' },
+                            { access: 'middleware-start', correlationId: '345678912', name: 'middleware3', startTime: '2016-02-29T14:54:41.073 -0800'},
+                            { access: 'middleware-end', correlationId: '345678912', name: 'middleware3', endTime: '2016-02-29T14:54:42.073 -0800', duration: 1000, result: 'end' },
```
avanderhoorn: Typically I try and give things reasonably real titles... it saves from everything being View1-5 and Action1-15. Any chance we can get a little more creative with the names?

src/request/components/request-detail-panel-execution.jsx
```diff
@@ -336,6 +338,60 @@ var ParamaterList = React.createClass({
...
+        
+        var content = null;
+        
+        if (middlewareStartMessages
+            && middlewareEndMessages
+            && middlewareStartMessages.length >= 0
+            && middlewareEndMessages.length >= 0) {
+        
+            middlewareStartMessages = middlewareStartMessages.sort(function (a,b) { return a.ordinal - b.ordinal; });
+            middlewareEndMessages = _.indexBy(middlewareEndMessages, 'payload.correlationId');
```
avanderhoorn: I know that `correlationId` made it into the schema, but I think that was a miss on my part from earlier. I really think this should be removed as we know its something that we are going to have trouble getting on other platforms. If you look at the way that `CommandList` does the correlation, you will notice that this just uses the `order` of the message and finds the next partner that matches. I think we should probably do the same here. Ultimately the logic which finds the next/matching partner can probably be abstracted out.

avanderhoorn:
I've gone through and made a couple of minor comments. Looking really good!

src/request/components/request-detail-panel-execution.jsx
philliphoff: I'm ok with that for now, but it makes assumptions that the enumerated values are designed for human consumption (as opposed to programmatic consumption) as well as English.  For localization purposes, such values are always explicitly mapped to a language-specific display value.

src/request/components/request-detail-panel-execution.jsx
philliphoff: Yes, I was just trying to follow the pattern/convention defined by the larger function.  I've removed it.

src/request/components/request-detail-panel-execution.jsx
philliphoff: I'll remove it.  That said, it's still unclear to me whether the action/view has any real relevancy in the Node world.  The "action" is really just the last piece of executed middleware; it doesn't seem to have any other real significance.  View, yes probably, for those applications doing server-side, template-based rendering.  However, according to @nebrius, that is actually the minority compared to applications doing client-side rendering.

fake/fake-extension.js
philliphoff: I had more representative names originally, then changed them when experimenting with sorting/ordering messages and just left them that way.  I've changed them back.

src/request/components/request-detail-panel-execution.jsx
avanderhoorn: Totally agree. Route is probably a more important item in this context. But for now we know this isn't the most important data item, and we can loop back around on what is.

src/request/components/request-detail-panel-execution.jsx
avanderhoorn: Totally agree. When it comes to that we will have a bit of work to do across the board for this.

fake/fake-extension.js
avanderhoorn: Great and makes sense.

src/request/components/request-detail-panel-execution.jsx
philliphoff: My concern is that it's not clear that the message ordinal *must* also correspond to the order in which the events occurred, across all of the platforms.  (Consider a case where the message publisher does the publishing off-thread for performance reasons; in that case it may be less likely for the ordinal has any bearing on message ordering within a given context.

I think we may still need some discussion around this to make sure we're doing the right thing.

avanderhoorn: It is a base assumption across the platform that order "ordinal must also correspond to the order in which the events occurred". We should probably try documenting that more. In the .net implementation the `order` is determined in the request to help with this problem. I'm happy to open the conversation around this again, but we already know we have concrete cases where a concept like `correlationId` in the message payload won't work. Beyond making sure its documented for platform authors, we know we can get `order` right. Does that make sense/sound reasonable or I might be missing something?

mike-kaufman:
lgtm

PR #50, opened 2016-03-30
[Middleware] Now displays middleware display name and package link, if available.

philliphoff:
Updates the Glimpse Client to display the middleware display name, if available.  If not, then the client will fallback to the middleware name, and finally, to `<anonymous>`.  The client will also display a link to the package to which the middleware belongs, if that is available.

screen shot 2016-03-30 at 1 51 27 pm

### DIFF
--- a/src/request/components/request-detail-panel-execution.jsx
+++ b/src/request/components/request-detail-panel-execution.jsx
@@ -577,10 +577,20 @@ var MiddlewareComponents = React.createClass({
                     case 'error': result = /* Exlamation Mark U+0021 */ String.fromCharCode(33); break;
                 }
 
+                var name = middlewareEndPayload.displayName || middlewareEndPayload.name || '<anonymous>';
+
+                var packageLink = null;
+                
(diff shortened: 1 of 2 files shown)


philliphoff:
Related to [#91](https://github.com/Glimpse/Glimpse.Node.Prototype/issues/91).

mike-kaufman:
lgtm.

src/request/components/request-detail-panel-execution.jsx
```diff
@@ -577,10 +577,20 @@ var MiddlewareComponents = React.createClass({
                     case 'error': result = /* Exlamation Mark U+0021 */ String.fromCharCode(33); break;
                 }
 
+                var name = middlewareEndPayload.displayName || middlewareEndPayload.name || '<anonymous>';
+
+                var packageLink = null;
+
```
avanderhoorn: nit for next time... some of these line breaks probably aren't necessary.

avanderhoorn:
:shipit:

PR #58, opened 2016-04-04
[Build] Create ZIP file of Client distribution files during CI build

philliphoff:
Updates the `ci` Gulp task to create a ZIP file "snapshot" of the distribution directory (minus the ZIP file itself, of course).  Also updates the Travis CI configuration to make the ZIP file part of a GitHub Release asset.  (This means that creating a release of the repo--i.e. tagging a commit--will cause Travis CI to build the repo, generate the ZIP file, then attach the ZIP file to the release for download).

This will then allow the Server and Electron repos to pull a given release snapshot during their build process, rather than have to actually copy the assets into their repositories.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,3 +5,10 @@ install:
 - npm install
 script:
 - gulp ci
+deploy:
+  provider: releases
+  api_key:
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -0,0 +1,14 @@
...
+node_js:
+- 5.6
+install:
+- npm install
+script:
+- gulp ci
+deploy:
+  provider: releases
+  api_key:
+    secure: p0DliyK4lDpmS26BGzvpxVh1UtEexDMSMZlokUyJvI+LqBfI8US3pCQDeo8kjAXhMTRJMlqJxm7oll8eyq6VMOt37t8juuBAfNh1WCcJBfojhpbcRPJ7PIhAn8djpJQgPRBYPhBXVU0HAyiryyab6AZdBXxRvEzFPonPwayvWEtJeJdjgBsMdL4TMDd3BkrWZhwCXlIGvRY5pJMb3QiNwLQlxHJ8s8K3emXn4IcPyKbg0jJy6uLTo4kDXFL/MOwB7etYpJKxNlQBcjaa7P8+Zf/xdmhwW3kYgKAO2JGZGF0lF6oQ5EZAbWlCERlOZg4cJmRjM0QovO/L3NXkWj4RzKy+9/5A6vitu3R4GkbRq7sOO7obPXXozSFZHdRzcswsEvI2IlM4KNwRnLQw170FfsgpvaP15nvNVnzIgXP9B/J/rAfyM/4FCPUixs9gEtt4SnhTyoo/Ty6jaCTjPNBvCQbed0Qgb0LjxTfwzTKNl/ScGN1VVp5wieZFrFrqCHj4/GjDEr4ywgApRucqGi7W+0IZO5UbFnG5ksqnwmg+V5tmv3/2AtQytOXqsE7QeMneflM3/L3y3+O18Q30Ocqlbx6R1R+hSjYzZ2vmau3SyI4Lx7cP2VgD2NR68Wz8GdaLyOX1yYEZa7BQ+uuJ3miPPvYc7b+iM0PQthgvlloVd5I=
```
avanderhoorn: Is this meant to be here?

gulpfile.js
```diff
@@ -117,10 +122,17 @@ gulp.task('dev', function (cb) {
...
 
     runSequence('build-prod', 'server', cb);
 });
 
+gulp.task('ci', ['build-dev'], function () {
+    return gulp.src(['dist/**', '!dist/client.zip'])
+        .pipe(zip('client.zip'))
+        .pipe(gulp.dest('dist'));
+});
+
```
avanderhoorn: Should we do a clean here? We don't really have a clean as part of the client build but probably not a bad idea at some point. That said, I'm assuming the CI starts with a clean slate each time?

avanderhoorn:
Only a minor questions but gtg assuming they are ok :shipit:

gulpfile.js
philliphoff: We should be starting with a clean slate on each build, but it probably wouldn't hurt to have a clean task anyway.  I can add that as a separate change.

avanderhoorn: Great. I'm happy for this to be a separate PR.

.travis.yml
philliphoff: The API key or the deployment configuration in general?  The latter is what causes a release to be generated (and ZIP file created) when a tag is attached to a commit.  The former is necessary for Travis to have access to the repo to do the attaching; the key itself is encrypted with a repo-specific key (held by Travis), so it's fine to be in the file.

avanderhoorn: I thought that was the case (re encryption) but wanted to check.

PR #59, opened 2016-04-04
Add clean step to CI build (to ensure clean ZIP files).

philliphoff:
Adds a `clean` Gulp task to delete the `dist` folder, and made it a dependency of the `ci` Gulp task.  This ensures that the `dist` folder is clean of any stray files that might otherwise make their way into a release ZIP file.


avanderhoorn:
Great!

PR #61, opened 2016-04-06
[Build] Include the HUD "package" with the CI and release process

philliphoff:
Adds support for packaging and releasing the HUD "package" in the same manner as the Client itself.


_temp/hud/package.json
```diff
@@ -26,21 +26,23 @@
...
+    "gulp-zip": "^3.2.0",
     "imports-loader": "0.6.5",
     "jscs": "2.3.4",
     "jshint": "2.8.0",
     "jshint-loader": "0.8.3",
     "lodash": "^3.10.1",
     "minimist": "1.2.0",
     "moment": "^2.10.6",
-    "node-sass": "3.3.3",
+    "node-sass": "3.4.2",
```
avanderhoorn: I remember that I had issues with updating to later versions of `node-sass`, but I don't know if that was for just the client or if I had the problem with HUD as well... do you know if things still work ok?

avanderhoorn:
One minor question about sass dependency update, but besides that, looks good :shipit:

_temp/hud/package.json
philliphoff: Travis didn't like the current version so I bumped it to the same version used by the client, so I wouldn't think there would be a problem.  If I pull in a newly-built version of the HUD into the Server, things still seem to work as expected.

avanderhoorn: Great!

PR #63, opened 2016-04-08
[Packing] Client fonts using file-loader.

philliphoff:
The Client is currently not able to find its custom fonts because WebPack is placing them relative to the root of the web application.  Using a relative path requires a custom loader (e.g. `file-loader`), which seems to resolve the problem.

(Interestingly, this same trick doesn't seem to work for the HUD, perhaps due to the way it's injected into the host page.)

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,6 +52,7 @@
     "del": "^2.2.0",
     "exports-loader": "0.6.2",
     "expose-loader": "0.7.1",
+    "file-loader": "^0.8.5",
     "gulp": "3.9.0",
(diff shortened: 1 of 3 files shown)


avanderhoorn:
Did this end up getting rid of all the 404s?

philliphoff:
Yes, no 404s; the requests for the fonts now return 200s.

philliphoff:
Feel free to give it a try before merging to make sure you're seeing what you expect.

avanderhoorn:
Once the conflicts here are sorted this can come in

PR #64, opened 2016-04-08
[Refactor] Add support for TypeScript-based React components

philliphoff:
Adds support for writing React components in TypeScript (i.e. TSX).  This allows one to fully define the types of properties passed between components and therefore get and compiler help with conforming to those types.

Also refactors the Request tab as an example of using TypeScript for React.


src/request/components/request-detail-panel-request.tsx
```diff
@@ -0,0 +1,90 @@
...
+                        <div className='tab-title col-10'>Url</div>
+                    </div>
+                    <div>{this.props.url}</div>
+                </div>
+            </div>
+        );
+    }
+}
+
+class RequestHeaders extends React.Component<{headers: { [key: string]: string }, title: string}, {}> {
```
avanderhoorn: Pity that the syntax here is a bit nicer :/ Should we create a header interface or something that says its an object of `{ [key: string]: string }`

avanderhoorn:
I've been wanting to move the client in this direction for a while now, just haven't had a chance to do it. So good work on kicking this off. Just a small comment about some of the type definitions.

src/request/components/request-detail-panel-request.tsx
```diff
@@ -0,0 +1,90 @@
...
+        'web-response': getItem,
+        'web-request': getItem
+    };
+
+    return function(request) {
+        return messageProcessor.getTypeStucture(request, options);
+    };
+})();
+
+class RequestUrl extends React.Component<{url: string}, {}> {
```
avanderhoorn: Does declaring the types like this buy us anything? Like it dosen't look like its a short cut for `this.props.url`... Does the type check come in the usage end with `<RequestUrl url={webRequestPayload.url} />`?

src/request/components/request-detail-panel-request.tsx
philliphoff: Yes, I should go back and define proper interfaces for the property bags passed to the component.  It would then look something like:

```javascript
export interface IRequestHeadersProperties {
    headers: { [key: string]: string };
    title: string;
}

class RequestHeaders extends React.Component<IRequestHeadersProps, {}> { }
```

I just did it the fastest way to remove the TypeScript compiler/linter errors.

src/request/components/request-detail-panel-request.tsx
philliphoff: Yes, once you declare the types for the properties passed to the component, you then get IntelliSense both when using those properties within the component as well as when setting the attributes on the component (from another component).  It works pretty slick.

philliphoff:
Resolving the conflicts may have to wait a bit, as it looks like the `typings` definition server is down so I can't verify a clean build.

src/request/components/request-detail-panel-request.tsx
avanderhoorn: Np, if we can do this, I think it would make a difference.

src/request/components/request-detail-panel-request.tsx
avanderhoorn: Sounds good!

avanderhoorn:
Not a problem.

PR #65, opened 2016-04-11
UX for logging levels and offsets

philliphoff:
Displays error/warning icons to trace message rows and displays the logging message offsets.

The icons are provided by [FontAwsome](http://fontawesome.io), by way of the React library [`react-fa`](https://github.com/andreypopp/react-fa).

### DIFF
--- a/src/request/components/request-detail-panel-logging.jsx
+++ b/src/request/components/request-detail-panel-logging.jsx
@@ -82,8 +82,8 @@ var LogMessages = React.createClass({
                         <th width="5%"><span className="table-col-title">#</span></th>
                         <th width="15%"><span className="table-col-title"><Icon fixedWidth="true" /> Level</span></th>
                         <th><span className="table-col-title">Message</span></th>
-                        <th width="10%"><span className="table-col-title">From Start</span></th>
-                        <th width="10%"><span className="table-col-title">Duration</span></th>
+                        <th width="15%"><span className="table-col-title">From Start (ms)</span></th>
+                        <th width="15%"><span className="table-col-title">Duration (ms)</span></th>
                     </tr>
                 </thead>
                 {this.props.logWriteMessages.map(function(message) {
@@ -95,7 +95,7 @@ var LogMessages = React.createClass({
                             <td>{payload.index}</td>
                             <td><Icon name={getIconName(payload.level)} fixedWidth="true" /> {getDisplayText(payload.level)}</td>
                             <td>{payload.message}</td>
-                            <td>-</td>
+                            <td>{message.context.offset}</td>
                             <td>-</td>
                         </tr>);
                 }) }
--- a/src/request/components/request-detail-panel-logging.jsx
+++ b/src/request/components/request-detail-panel-logging.jsx
@@ -5,6 +5,7 @@ var messageProcessor = require('../util/request-message-processor');
 var moment = require('moment');
 var _ = require('lodash');
 var React = require('react');
+var Icon = require('react-fa');
 
 /**
  * Return the messages to be used by the view.
@@ -45,6 +46,30 @@ function getRowClass(message) {
     return rowClass;
 }
 
+function getDisplayText(level) {
+    return _.startCase(level);
+}
+
+function getIconName(level) {
+    switch (level.toLowerCase()) {
+        case 'critical':
+        case 'error':
+            return 'times-circle';
+
+        case 'warning':
+            return 'exclamation-triangle';
+
+        case 'verbose':
+        case 'information':
+        default:
+            return '';
+    }
+}
+
+function getMessagesDisplayText(messages) {
+    return messages.length + ((messages.length === 1) ? ' Message' : ' Messages');
+}
+
 /**
  * React class to display console messages
  */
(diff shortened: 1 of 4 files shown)


src/request/components/request-detail-panel-logging.jsx
```diff
@@ -26,16 +27,16 @@ var getMessages = (function() {
  */
 function getRowClass(message) {
```
avanderhoorn: If we are doing what we are doing in `getIconName

src/request/components/request-detail-panel-logging.jsx
```diff
@@ -45,6 +46,30 @@ function getRowClass(message) {
     return rowClass;
 }
 
+function getDisplayText(level) {
+    return _.startCase(level);
+}
+
+function getIconName(level) {
```
avanderhoorn: Originally the intent was to use the class defined in `getRowClass` to load in the corresponding svg icon. I think this would still be my preferred approach than loading in a whole new font. But I'm happy to pull this in in the short term.

src/request/components/request-detail-panel-logging.jsx
```diff
@@ -55,22 +80,22 @@ var LogMessages = React.createClass({
                 <thead>
                     <tr className="table-col-title-group">
                         <th width="5%"><span className="table-col-title">#</span></th>
-                        <th width="10%"><span className="table-col-title">Level</span></th>
+                        <th width="15%"><span className="table-col-title"><Icon fixedWidth="true" /> Level</span></th>
                         <th><span className="table-col-title">Message</span></th>
-                        <th width="10%"><span className="table-col-title">From Start</span></th>
-                        <th width="10%"><span className="table-col-title">Duration</span></th>
+                        <th width="15%"><span className="table-col-title">From Start (ms)</span></th>
+                        <th width="15%"><span className="table-col-title">Duration (ms)</span></th>
```
avanderhoorn: I think the final design dropped the `(ms)`... @chetsangnil Do you have any thoughts on `(ms)` vs not?

src/request/components/request-detail-panel-logging.jsx
philliphoff: To a certain extent, it was an experiment in how easy it would be to incorporate such a library.

Personally, I'd lean towards using a library of web-wide common iconography that does all of the work for me (and makes it easy to manipulate images w.r.t. size, orientation, layout, etc.) rather than create, manage, and manipulate individual, custom images.

src/request/components/request-detail-panel-logging.jsx
philliphoff: The CSS is still applying certain properties (e.g. foreground color), so it's still needed.  If it's possible and reasonable (and that's just showing my ignorance of CSS) to put that all in CSS, than that's fine with me, too.

src/request/components/request-detail-panel-logging.jsx
philliphoff: IMO, a diagnostics tools should *always* indicate its units to disambiguate comparisons between values.  I don't have a strong preference for whether that's at a column level (which tends to reduce clutter) or at a value level (which tends to improve readability and understanding).  (The same goes for timestamps, which should always indicate whether a time is local or UTC.)

src/request/components/request-detail-panel-logging.jsx
avanderhoorn: I agree with you, but we have been supplied with images we are meant to use for various reasons. I've got a copy of them which I can provide.

src/request/components/request-detail-panel-logging.jsx
avanderhoorn: I like it to, but it was removed. Hence why I wanted input from Chet as to if this was for a specific reason.

src/request/components/request-detail-panel-logging.jsx
avanderhoorn: This will probably be needed anyway given that we are using our own icons.

chetsangnil:
Sorry for the confusion guys, the unit "ms" should be at the value level!! I must have missed it in the mock!!

mike-kaufman:
lgtm

chetsangnil:
Like so!

glimpse v_8

chetsangnil:
and the reason why I chose to show "ms" at the value level is because 1.) Phil was right on the spot about improve readability and understanding 2.) numbers within the column will generally be shorter than its title. If we add "ms" next to the title of the column, we would run into a lot of empty space within the column itself.

But yes! please put back in the unit. It's very important!

chetsangnil:
@avanderhoorn : if you look at the mock for Web services tab, you will see that "ms" is shown inside the column without a space--- (ex. 123ms). To keep it consistent with the logging page, please update for it to show with a space--- (ex. 123 ms). Thank you!

chetsangnil:
About font, font awesome is great and I think we can definitely leverage the use of that. However, @avanderhoorn is right that we are still going to need our own set of icon as well because font awesome will not have all the assets we need to show in the UI.

philliphoff:
Abandoning this PR in favor of a more recent attempt.

PR #66, opened 2016-04-13
[Refactor] Logging tab to TypeScript and V-VM

philliphoff:
Refactors the logging tab in two primary ways: 
- Ports the JSX to (TypeScript) TSX.
- Splits out the logic for building a "model" for the view into a separate component (a view model).


src/request/view-models/LoggingViewModel.ts
```diff
@@ -0,0 +1,38 @@
...
+
+    private static options = {
+        'log-write': LoggingViewModel.getList
+    };
+
+    private _messages: ILogMessage[];
+
+    public get messages() {
+        return this._messages;
+    }
```
avanderhoorn: I'm assuming that you are doing this as a prop is it immutable?

src/request/messages/ILogMessage.ts
```diff
@@ -0,0 +1,6 @@
+'use strict';
+
+export interface ILogMessage {
+    level: string;
+    message: string;
+}
```
avanderhoorn: Will be interesting to see what we end up doing with contracts like this as they are more of a mixin rather than a single object.

src/request/components/request-detail-panel-logging.tsx
```diff
@@ -0,0 +1,100 @@
...
+    request;
+    viewModel: ILoggingViewModel;
+}
+
+/**
+ * React class to for the console log messages tab
+ */
+export class Logging extends React.Component<ILoggingProps, {}> {
+    public getInitialState() {
+        return { checkedState: false };
```
avanderhoorn: I know this was brought across from the old file, but can you see any reason why this is here?

src/request/components/request-detail-panel-logging.tsx
```diff
@@ -0,0 +1,100 @@
...
+import React = require('react');
+
+interface ILogMessagesProps {
+    messages: ILogMessage[];
+}
+
+/**
+ * React class to display console messages
+ */
+class LogMessages extends React.Component<ILogMessagesProps, {}> {
```
avanderhoorn: Now that this is simplified, I wonder if there is any benefit to having this in a separate component?

src/request/components/request-detail-panel-logging.tsx
```diff
@@ -0,0 +1,100 @@
...
+/**
+ * React class to for the console log messages tab
+ */
+export class Logging extends React.Component<ILoggingProps, {}> {
+    public getInitialState() {
+        return { checkedState: false };
+    }
+
+    public render() {
+        if (!_.isEmpty(this.props.viewModel.messages)) {
```
avanderhoorn: This pattern where we check to see if we have anything and if now we display a message can probably be abstracted into a shared component - https://github.com/Glimpse/Glimpse.Client.Prototype/tree/master/src/lib/components. That said, thats outside the scope of this pr.

src/request/request-tab.js
```diff
@@ -34,7 +34,8 @@ module.exports = {
 // TODO: Need to come up with a better self registration process
 require('./components/request-detail-panel-execution');
 var request = require('./components/request-detail-panel-request');
-require('./components/request-detail-panel-logging');
+var LoggingComponent = require('./components/request-detail-panel-logging').Logging;
+var LoggingViewModel = require('./view-models/LoggingViewModel').LoggingViewModel;
```
avanderhoorn: I wonder if the models should be put in a folder under `components` called `models`. That way it implies that the models are for the component. Plus they don't really have a use outside of the components they apply to.

src/request/request-tab.js
```diff
@@ -43,3 +44,10 @@ module.exports.registerTab({
     title: 'Request',
     component: request.Request
 });
+
+module.exports.registerTab({
+    key: 'tab.logging',
+    title: 'Trace',
+    component: LoggingComponent,
+    viewModel: LoggingViewModel
```
avanderhoorn: I wonder, to make the registrations a little cleaner, if the requires should just be used in place or if we want to put them into a var, just above the matching `registerTab`.

avanderhoorn:
Ok, now that the conversion is done, any thoughts on how this looks, if you think it has simplified things, how it will play out across other tabs, etc?

src/request/messages/ILogMessage.ts
philliphoff: In the case where components expect the payloads to be a mix of message types, they can always define a composite interface that aggregates the properties of each type.  In the case where a component might need to treat the payload as completely opaque, it can use the non-generic `IMessage` type.  We'll just have to see how well that works out in practice.

src/request/components/request-detail-panel-logging.tsx
philliphoff: You mean the `checkedState: false`?  I'm not sure what it does, to be honest.

src/request/components/request-detail-panel-logging.tsx
philliphoff: If it were re-used by other components or was unit tested independently, definitely.  If used only by this component, and being a view component where no unit tests are generally expected, then I'd tend to treat this like a "private class" in .NET.  It's there for the purposes of making the larger component easier to read, but is really just an implementation detail not meant to be visible outside that larger component.

src/request/components/request-detail-panel-logging.tsx
philliphoff: I'd say that depends on how similar the "no data" UX is between components, and the similarity in determining that fact between view models, such that the make abstraction ultimately saves time or maintenance.  It's definitely been done--we used that technique on my last team which had a very consistent UX across a number of components in both loading/error/no-data cases.  But, I'd wait until we saw a number of components which needed that same UX first.

src/request/messages/ILogMessage.ts
avanderhoorn: That was my assumption.

src/request/components/request-detail-panel-logging.tsx
avanderhoorn: Yep. maybe @mike-kaufman might know what its for.

src/request/request-tab.js
philliphoff: I've always preferred to treat views and view-models as peers rather than imply a hierarchy or "owner".  They are used outside of the view--for testing!  The views should care only that what's given to them meets the agreed-upon contract for property binding, not what or how or by whom those contracts are implemented.

Obviously the file hierarchy doesn't necessarily have any bearing on that, but I find that keeping them as peers tends to prevent the thinking that one serves the other, which might then skew the design in that, undesired, direction. 

At the end of the day, though, it's not that big of a deal.

src/request/request-tab.js
philliphoff: Yeah, it's still a little messy.  I'll move things around to be a little more clear.

src/request/components/request-detail-panel-logging.tsx
avanderhoorn: In this case because there is no reuse else where and the complexity is low, it would probably be simpler to read if it was just one component but its not a big deal.

src/request/view-models/LoggingViewModel.ts
philliphoff: The idea was that each instance of the `Logging` component get its own instance of the view model, which is initialized before component creation by the (same) request.  The message data held by the view model from that point should be effectively immutable (assuming my understanding of how/when React creates and renders components).  

Obviously things will change once filtering is involved, as those user actions will have notify the view model to sort/filter those messages in different ways, then the component will need to re-render, binding to the changes.

src/request/request-tab.js
avanderhoorn: I get you point. I guess part of my feeling is that this are models for the "components" so if anything the folder should be "component-models" but that seems a little long winded. Also the fact that besides in testing, no other sub system besides the components use these models. Given this, if they sit as a peer, it almost feels like then its a peer of the rest of the sub systems which isn't really the case. Does that make any sense?

src/request/components/request-detail-panel-logging.tsx
avanderhoorn: Agreed. Will bring it up with @chetsangnil at some point.

src/request/components/request-detail-panel-logging.tsx
mike-kaufman: I think that's something I just copy/pasted from the request-detail-panel-execution tab:  https://github.com/Glimpse/Glimpse.Client.Prototype/blob/master/src/request/components/request-detail-panel-execution.jsx#l673.

If not used, it can be removed.

src/request/request-tab.js
philliphoff: I'll concede the point about naming--it's unfortunate that React uses the terribly-generic term "component" for what everyone else would call a "view".  I'd be fine renaming the folder to `component-model` and types `IComponentModel`-based.

I disagree that the view models are *for* the views, as opposed to a component in its own right that exists to encapsulate and coordinate the interaction between the user, the application, and its data.    It's in that sense that I prefer to keep view models as peers of the views.  (Whether both should be at a peer level with the rest of the application components is another question.)

avanderhoorn: Ok I can buy that.

src/request/components/request-detail-panel-logging.tsx
philliphoff: Oh, I see what you mean.  The two components were just a hold-over from the original.  I've merged it into a single component which does eliminate a little React boilerplate.

philliphoff:
>Ok, now that the conversion is done, any thoughts on how this looks, if you think it has simplified things, how it will play out across other tabs, etc?

I think it will simplify things, especially once the view model has some real behavior (such as filtering).  Of  course, it's going to take doing a more complicated tab and/or feature within the logging tab to be sure.  And probably a conversion of a third tab to see what can be further abstracted or consolidated within a base view model.

avanderhoorn:
:shipit: Great work mate

PR #67, opened 2016-04-14
[HttpClient] Initial tab implementation

avanderhoorn:
This is the initial cut of the HttpClient tab. Its certainly not complete yet (as none of the headers are being processed yet and we are missing timing information), but given where we are at currently this will have to do.

screen shot 2016-04-13 at 9 07 41 pm


src/request/components/request-detail-panel-webservices.jsx
```diff
@@ -0,0 +1,190 @@
...
+        
+        return (
+            <table className="table table-bordered table-striped tab-content-item table-selectable">
+                <thead>
+                    <tr className="table-col-title-group">
+                        <th><span className="table-col-title">Name/Path</span></th>
+                        <th width="10%"><span className="table-col-title">Status</span></th>
+                        <th width="10%"><span className="table-col-title">Method</span></th>
+                        <th width="10%"><span className="table-col-title">Protocal</span></th>
+                        <th width="10%"><span className="table-col-title">Type</span></th>
```
philliphoff: Protocol?

src/request/components/request-detail-panel-webservices.jsx
```diff
@@ -0,0 +1,190 @@
...
+            // TODO: replace with correlation matching once inplace
+            var httpRequest = dataHttpRequestMessages[i].payload;
+            var httpResponse = dataHttpResponseMessages[i].payload;
+            
+            requestItems.push(
+                <tr onClick={this.props.onSelectedIndex.bind(this, i)} className={classNames({
+                        'selected': i == this.props.selectedIndex
+                    })}> 
+                    <td>{httpRequest.url}</td>
+                    <td>{httpResponse.statusCode}</td>
```
philliphoff: Should the HTTP method be uppercased (to be consistent with standard usage)?

src/request/components/request-detail-panel-webservices.jsx
```diff
@@ -0,0 +1,190 @@
...
+        
+        // get child items
+        var requestItems = [];
+        for (var i = 0; i < dataHttpRequestMessages.length; i++) {
+            // TODO: replace with correlation matching once inplace
+            var httpRequest = dataHttpRequestMessages[i].payload;
+            var httpResponse = dataHttpResponseMessages[i].payload;
+            
+            requestItems.push(
+                <tr onClick={this.props.onSelectedIndex.bind(this, i)} className={classNames({
```
philliphoff: Should the items in most of these columns be centered?  Status code certainly, probably method and protocol.

src/request/components/request-detail-panel-webservices.jsx
```diff
@@ -0,0 +1,190 @@
...
+        
+        var content = null;
+        if (dataHttpRequestMessages && dataHttpResponseMessages) {
+            var selectedDataHttpRequestMessage = dataHttpRequestMessages[this.state.selectedIndex];
+            var selectedDataHttpResponseMessage = dataHttpResponseMessages[this.state.selectedIndex];
+            
+            // intial processing of messages
+            dataHttpRequestMessages = dataHttpRequestMessages.sort(function(a, b) { return a.ordinal - b.ordinal; });
+            dataHttpResponseMessages = dataHttpResponseMessages.sort(function(a, b) { return a.ordinal - b.ordinal; });
+
```
philliphoff: Making some assumptions here that the number of request/response messages are equal and that the pairs are strictly sequential; if a component makes several requests in parallel, this may not hold up.  But that goes back to our earlier discussions about correlation IDs.  It's probably fine for the purposes of the demo (as I don't think the sample app makes multiple external requests in parallel).

src/request/components/request-detail-panel-webservices.jsx
```diff
@@ -0,0 +1,190 @@
...
+    return function(request) {
+		return messageProcessor.getTypeStucture(request, options); 
+    }
+})();
+
+/**
+ * React class to display console messages
+ */
+var ServiceMessages = React.createClass({
+    _changeSelected: function(index) {
```
philliphoff: Is this used anywhere (e.g. React)?

src/request/components/request-detail-panel-webservices.jsx
avanderhoorn: Good catch

philliphoff:
Other than my comments, LGTM.

src/request/components/request-detail-panel-webservices.jsx
avanderhoorn: Good question. At the moment I'm showing what ever has come from the server. What formatting we want to put around that should be brought up. Given timing that is probably out of the scope of this PR.

src/request/components/request-detail-panel-webservices.jsx
avanderhoorn: Not according to the design. We can bring that up post this PR.

src/request/components/request-detail-panel-webservices.jsx
avanderhoorn: Totally agree, hence why we are introducing the correlation ids. That will be part of the cleanup next sprint.

src/request/components/request-detail-panel-webservices.jsx
avanderhoorn: I thought I had removed it. Good catch.

PR #71, opened 2016-04-25
Add test infrastructure to Glimpse client

jasonyueyang:
We build and run test cases, written in TypeScript, using Mocha and Chai.

We use Webpack to collect, transpile these tests, and put them under `dist-test`.

Run `npm test` to launch testing.

This update includes:
- Gulp tasks `clean-test`, `build-test`, and `test`
- Mocha, Chai, and their typings info
- Instruction in README
- A dummy repository test

Currently test collection is hardcoded. This will be changed in future updates.


gulpfile.js
```diff
@@ -118,6 +128,33 @@ gulp.task('build-ci', ['clean'], function (cb) {
     runSequence('build', cb);
 });
 
+gulp.task('build-test', function buildTest(cb) {
+    var config = _.defaultsDeep({}, require('./webpack.config'));
+
+    config.entry = {};
+    config.entry[path.join('request', 'repository', 'request-repository-cache')] = './test/request/repository/request-repository-cache.spec.ts';
```
nebrius: Can you add a comment here mentioning that the `path.join` bit for the key in the entry causes the entry bundle to be put in that folder structure?

typings.json
```diff
@@ -1,5 +1,7 @@
 {
   "ambientDependencies": {
+    "chai": "github:DefinitelyTyped/DefinitelyTyped/chai/chai.d.ts#a9680ac2efd3add0d38cd0331f30ad497d9dd1e0",
+    "mocha": "github:DefinitelyTyped/DefinitelyTyped/mocha/mocha.d.ts#a9680ac2efd3add0d38cd0331f30ad497d9dd1e0",
```
nebrius: I'm curious about this one. I don't see mocha being imported anywhere (which is expected), but it's also not manually included in the build process. Does this mean that this is always injected? And does that mean that I could do `describe()` in client code, and it wouldn't error? Do we care if that's the case?

gulpfile.js
jasonyueyang: Done.

nebrius:
:shipit:

typings.json
jasonyueyang: You're probably right that our current typings setup would treat these Mocha function types as "global". Separating type definitions between client and test is feasible, but seems to be an overkill at this point.

gulpfile.js
```diff
@@ -118,6 +128,34 @@ gulp.task('build-ci', ['clean'], function (cb) {
...
+
+        if (!started) {
+            started = true;
+            cb();
+        }
+    }
+
+    var compiler = webpack(config);
+    compiler.run(processResult);
+});
```
avanderhoorn: This runs the whole `webpack` build... I know we talked about the need to inject fake/mock modules. Are we looking at dealing with that yet or are we going to bring that online later. Note, I know we change the `config.entry` point, but how are we mocking these. 

Additionally, at the moment it looks like we need to manually target specific `*.spec`s, is this just temporary or are we looking to change this later on.

jasonyueyang: When we get to the point where we need to mock modules, I'm planning to use something like mockery.

Regarding manually connecting tests, it will be changed by my next PR.

avanderhoorn: Sounds great. Let get this in!

PR #72, opened 2016-04-26
Automate the process of collecting tests

jasonyueyang:
Used `walk` to automatically collect all test cases. 

Added another dummy test for stores.


gulpfile.js
```diff
@@ -58,6 +60,21 @@ function getBundleConfig() {
     return config;
 }
 
+function getTests(cb) {
+    var walk = require('walk');
```
nebrius: Can we move this to the top of the file with the other require?

gulpfile.js
```diff
@@ -129,26 +146,35 @@ gulp.task('build-ci', ['clean'], function (cb) {
...
-    function processResult(err, stats) {
-        gutil.log('Webpack\n' + stats.toString(config.log));
-
-        if (!started) {
-            started = true;
-            cb();
+    getTests(function collectAllTests(tests) {
+        var config = _.defaultsDeep({}, require('./webpack.config'));
+        var pathPrefix = path.join(testSettings.input, '/');
+        var re = new RegExp(testSettings.testSuffix + '$');
```
nebrius: Can we move this logic that filters the tests into the `getTests` method? Also, can we have testSuffix itself be a regex instead of constructing it here, e.g.

```
var testSettings = {
    ...
    testSuffix: /\.spec.ts$/
  };
```

gulpfile.js
```diff
@@ -129,26 +146,35 @@ gulp.task('build-ci', ['clean'], function (cb) {
...
-    var started = false;
-    function processResult(err, stats) {
-        gutil.log('Webpack\n' + stats.toString(config.log));
-
-        if (!started) {
-            started = true;
-            cb();
+    getTests(function collectAllTests(tests) {
+        var config = _.defaultsDeep({}, require('./webpack.config'));
+        var pathPrefix = path.join(testSettings.input, '/');
```
nebrius: can we use `path.sep` instead of `/` here to make absolutely sure it works on Windows?

gulpfile.js
```diff
@@ -129,26 +146,35 @@ gulp.task('build-ci', ['clean'], function (cb) {
...
+        var re = new RegExp(testSettings.testSuffix + '$');
+
+        config.entry = {};
+        tests.forEach(function collectTest(test) {
+            var pathName = test.replace(pathPrefix, '');
```
nebrius: There's a helper method in `path` called `path.relative` that's a little more elegant that can be used here instead of doing a string replacement.

gulpfile.js
jasonyueyang: 👍

gulpfile.js
jasonyueyang: 👍 

Using `'/'` in `path.join` does work on Windows, but `path.sep` is cleaner.

nebrius: well, here's the thing, it _usually_ works on Windows, but it depends on the context of how it's being used. If it's being used with built-in Node.js libraries, it always works. If it's with 3rd party libraries, it _almost_ always works, but if you use it in anything that does an `exec` or `spawn`, it _never_ works.

gulpfile.js
jasonyueyang: Converted testSuffix to regex. 

Kept the logic in the same place since `getTests` returns the array of full path names and here we simply strip out the suffix.

gulpfile.js
jasonyueyang: Here we want to strip out the prefix from the full path name. Replacing the prefix with `''` should be fine. I added some comments to make the intent more clear.

gulpfile.js
```diff
@@ -129,26 +146,38 @@ gulp.task('build-ci', ['clean'], function (cb) {
...
-    function processResult(err, stats) {
-        gutil.log('Webpack\n' + stats.toString(config.log));
-
-        if (!started) {
-            started = true;
-            cb();
+    getTests(function collectAllTests(tests) {
+        var config = _.defaultsDeep({}, require('./webpack.config'));
+        var pathPrefix = path.join(testSettings.input, path.sep);
+        var re = new RegExp(testSettings.testSuffix);
```
nebrius: `testSuffix` is already a regex, so there's no need to recreate it here, you can just use it directly.

gulpfile.js
```diff
@@ -58,6 +61,20 @@ function getBundleConfig() {
     return config;
 }
 
+function getTests(cb) {
+    var files = [];
+    var walker = walk.walk(testSettings.input, { followLinks: false });
+
+    walker.on('file', function(root, stat, next) {
+        files.push(path.join(root, stat.name));
```
nebrius: Sicne we removed the filter from the gulp task, can we add it back here? We do still need to filter out anything that's not a `spec.js` file.

gulpfile.js
jasonyueyang: Good point.

gulpfile.js
jasonyueyang: 👍

gulpfile.js
jasonyueyang: Good to know.

nebrius:
:shipit:

gulpfile.js
```diff
@@ -129,26 +148,37 @@ gulp.task('build-ci', ['clean'], function (cb) {
...
+
+        var started = false;
+        function processResult(err, stats) {
+            gutil.log('Webpack\n' + stats.toString(config.log));
+
+            if (!started) {
+                started = true;
+                cb();
+            }
         }
```
avanderhoorn: Does `getTests` ultimately get called once per test file?

gulpfile.js
```diff
@@ -14,6 +14,7 @@ var runSequence = require('run-sequence');
 var zip = require('gulp-zip');
 var mocha = require('gulp-mocha');
 var path = require('path');
+var walk = require('walk');
```
avanderhoorn: Out of curiosity, can `walk` be used in "non test" scenarios to "discover" and require modules?

avanderhoorn:
Good work guys on seeing this home! So glad to see that we are getting the test infrastructure in place. Is the test reporting tied into our CI build yet? and we should probably look at coverage reports at some point (get a better sense on how step the hill is).

gulpfile.js
nebrius: Definitely. It's a generic tool for walking directory trees and could be used for just about anything.

PR #76, opened 2016-04-27
Display logging offsets, if available.

philliphoff:
If logging messages are published with corresponding offsets, displays the offsets on the logging tab.

### DIFF
--- a/fake/fake-extension.js
+++ b/fake/fake-extension.js
@@ -787,7 +787,7 @@ var generateMvcRequest = (function() {
         },
         createLog: function(log, context) { 
             var message = this.createMessage('log-write', context);
-            mapProperties(log, message.payload, [ 'replacedRegions', 'message', 'level' ]);
+            mapProperties(log, message.payload, [ 'replacedRegions', 'message', 'level', 'offset' ]);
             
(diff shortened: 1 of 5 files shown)


avanderhoorn:
Looks good. Only thing is that the offset will likely need to be computed from the timestamp. My memory of the discussion we had was that all the messages would be coming in with a normalized timestamp but we would still need to calculate the offset locally.

philliphoff:
> Looks good. Only thing is that the offset will likely need to be computed from the timestamp. My memory of the discussion we had was that all the messages would be coming in with a normalized timestamp but we would still need to calculate the offset locally.

Based on @mike-kaufman's summary email, the decision was for the offset to be included as message payload and belonging to an offset-specific message type (where the offset is relative to the timestamp that represents the beginning of the associated context).  The fact that that context timestamp will (eventually) be generated by a clock sync'd with others (making offsets more comparable across contexts and across agents), is a separate question.

mike-kaufman:
Is this ready to go in?  Been sitting here for a while...

philliphoff:
@mike-kaufman: There's still ongoing discussion about what message type the offset property should belong to, as well as a couple of other things, so this PR is shelved for now.

avanderhoorn:
Closing for the time being.

PR #82, opened 2016-05-04
Updated all deps except React

nebrius:
This updates all dependencies except React and React related dependencies. It also introduces npm-shrinkwrap to update dependences and lock down the dependency tree.

The two biggest things in this PR are the addition of shrinkwrap and updating lodash to version 4 (this required a hand audit of the codebase and accounts for most of the changes).

To install the dependencies, run `npm install` like you normally would and it will automatically use the shrinkwrapped dependency tree found in `npm-shrinkwrap.json`. If you want to update a dependency, or add a new dependency, make your change to `package.json` and then run `gulp update-deps`.


nebrius:
Note: I also found a couple of places with inconsistent whitespace, so I went ahead and updated those too (extra trailing whitespace, using tabs instead of space, things like that). I only updated things I ran across though, so there may still be some inconsistent white space.

avanderhoorn:
Looks great to me. Just going to hold off bringing in so that @philliphoff get a chance to see that he will need to `npm install` to get up and running.

nebrius:
Along those lines, I was thinking it might be useful to have the build check it see if dependencies are out of date and do an `npm install` if needed, or heck just do it automatically since installs are fast if there's nothing to do. Thoughts?

philliphoff:
> Looks great to me. Just going to hold off bringing in so that @philliphoff get a chance to see that he will need to npm install to get up and running.

Looks ok to me.  It will be nice to be on the latest Lodash release.

nebrius:
Thought more about it, and we can add more automation around `npm install` later. Let's try this out as is for now and go from there.

avanderhoorn:
@nebrius I like the idea of doing it automatically... it would be nice to let the user know though if an update occurs. Regardless of this, I don't think its a blocker for this PR. Lets :shipit:

Issue #78, opened 2016-04-29
[Logging] Filtering by level


philliphoff:
Allow users to select only the messages in which they're interested, by logging level (e.g. all, info, warning, error, etc.).

philliphoff:
Linked to #46.

PR #83, opened 2016-05-05
Adds filtering of trace/log messages

philliphoff:
Allows the user to filter trace/log messages by level.

screen shot 2016-05-05 at 8 47 16 am

The "component model" (i.e. "view model") concept has been split into separate components, the model itself and optional state which can reside with the React component (UI) and persist across models.  (This enables the filter selection to be "sticky" as the user switches between requests, making comparisons of logs easier.)

The styling of the tab has also been updated to match the latest redlines.


philliphoff:
Related to #78.

src/request/components/request-detail-panel-logging.tsx
```diff
@@ -1,49 +1,67 @@
...
+                                    <tr className='tab-logs-data-default' key={message.id}>
+                                        <td>{message.ordinal}</td>
+                                        <td className={Logging.getRowClass(message)}>{message.level}</td>
+                                        <td>{message.message}</td>
+                                        <td>-</td>
+                                        <td>-</td>
+                                    </tr>);
+                            })
+                        }
+                        </tbody>
```
avanderhoorn: Shouldn't this `tbody` be between the `thead` and the `tfoot`?

src/shell/components/shell-view.scss
```diff
@@ -11,6 +11,11 @@
     src: url('../../../assets/selawkl.woff2') format('woff2'),
          url('../../../assets/selawkl.woff') format('woff');
 }
+@font-face {
+    font-family: 'Selawik Semibold';
+    src: url('../../../assets/selawkl.woff2') format('woff2'),
+         url('../../../assets/selawkl.woff') format('woff');
+}
```
avanderhoorn: Is there much point to having this if its the exact same asset as `Selawik Light`? Or is the intent to get the `Semibold` font as well?

src/shell/components/shell-view.scss
```diff
@@ -57,9 +62,9 @@ h1, h2, h3, h4, h5, h6 {
 
 
 //****** TABLE *******
-$table-border-color: #dbdbdb;
+$table-border-color: #ccc;
 $table-border-width: 1px;
-$table-background-alt-color: #fafafa;
+$table-background-alt-color: #f9f9f9;
```
avanderhoorn: I'm assuming these color updates came from Chet? Not that it really matters, just getting up to speed.

src/shell/components/shell-view.scss
```diff
@@ -97,12 +102,7 @@ $table-background-alt-color: #fafafa;
     thead {
         .table-col-title-group th {
             border-bottom: $table-border-width solid $table-border-color;
-        }
-        
-        th:first-child .table-col-title,
-        .table-col-featured .table-col-title {
-            border-left: 0px;
-        }
+        }
```
avanderhoorn: Why was this removed? It prevents the left side boarder from showing in the first column.

src/request/components/request-detail-panel-logging.tsx
philliphoff: According to the [documentation](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/tbody) (and the only way I could get everything styled as expected), it's supposed to come *after* the `thead` and `tfoot`.  I suspect that browsers simply deal with the fact that many developers do not do that.  (The spec is definitely not intuitive, but probably done that way for layout-while-downloading reasons.)

src/shell/components/shell-view.scss
philliphoff: It's to get the `Semibold` font as well, according to @chetsangnil.

src/shell/components/shell-view.scss
philliphoff: Yes, based on the latest redlines.  He updated some of the table-related colors across the board.

src/shell/components/shell-view.scss
philliphoff: @chetsangnil explicitly wants that left border on the first column of tables.

src/request/components/request-detail-panel-logging.tsx
avanderhoorn: Looks like come "html5" you can put the tfoot before, but prior to html5 it had to go between it - http://stackoverflow.com/questions/18901950/thead-tfoot-and-tbody-order-in-html5. What was the weird behavior you where seeing?

src/shell/components/shell-view.scss
avanderhoorn: But at the moment it pointing to the same asset as the `Selawik Light`...

src/shell/components/shell-view.scss
avanderhoorn: Really! It wasn't on the earlier iterations.

src/request/components/request-detail-panel-logging.tsx
philliphoff: I *recall* that it had something to do with the keys attached to each row (needed by React to properly add/remove items).  I can move the `tbody` around and see if it was just a fluke of something else I was doing at the time.

avanderhoorn: It would be nice to shift it back up as it kinda does make more sense to have it above, but don't spend tons of time pon it.

src/shell/components/shell-view.scss
avanderhoorn: Chatted with Chet, this is fine for now.

src/shell/components/shell-view.scss
philliphoff: Ack, you're right.  I'll work with @chetsangnil  to get the `sb` version of the font.

src/request/component-models/LoggingComponentModel.ts
```diff
@@ -1,39 +1,164 @@
...
+        return this._message.payload.level;
+    }
+
+    public get message(): string {
+        return this._message.payload.message;
+    }
 
-    private static options = {
-        'log-write': LoggingComponentModel.getList
-    };
```
avanderhoorn: Any particular reason this got pulled from being a static to inside the init method?

avanderhoorn: I personally find it a little easier to follow with the way you have it now, but just wondering in general.

philliphoff: In order to inject the message processor for unit testing, those fields could no longer be static (as they referred to the message processor).

src/request/components/request-detail-panel-logging.tsx
philliphoff: Seems to work ok now that I've moved it back.

src/shell/components/shell-view.scss
philliphoff: I've added the semibold (and semilight) font assets.

src/shell/components/shell-view.scss
philliphoff: As directed by @chetsangnil, I've switch from using a left border to a right border.

fake/fake-extension.js
```diff
@@ -885,9 +887,10 @@ var generateMvcRequest = (function() {
 
             // child actions
             if (action.actions) {
+                var that = this;
                 _.forEach(action.actions, function(childAction) {
-                    this.processChildActions(childAction, request, context);
-                }, this);
+                    that.processChildActions(childAction, request, context);
+                });
```
avanderhoorn: Just wondering why the switch from this to that... just cleaning things up?

src/request/component-models/ComponentModel.ts
```diff
@@ -0,0 +1,7 @@
+'use strict';
+
+import { IComponentModel } from './IComponentModel';
+
+export abstract class ComponentModel implements IComponentModel {
+    public abstract init(request);
+}
```
avanderhoorn: Is it still worth having this instead of just implementing the interface directly.

fake/fake-extension.js
philliphoff: Yeah, the Lodash `_.forEach()` changed its API in a backward-incompatible way from 3.x to 4.x, where you can no longer pass in the `this` object with which to bind.  This was breaking the Client fakes.

src/request/component-models/ComponentModel.ts
philliphoff: Since it was already there, I kept it "just in case" there was a future need to have some base functionality for each component model.

src/request/stores/request-detail-store.js
```diff
@@ -102,3 +105,18 @@ function getTabs(request) {
...
+// Toggle Logging Filter Level
+(function () {
+    function triggerRequest(payload) {
+        _viewModel.logging.filter = payload;
+
+        requestChanged(_viewModel);
+    }
+
+    glimpse.on('data.request.detail.logging.filter', triggerRequest);
+})();
```
avanderhoorn: Pity that this needs to happen so "high" up... I wonder if it would be better if the component model itself should listen for this event and emit an event that the tab is listen for. The problem with this approach though is that we are short circuiting the update chain, but I'm on the fence as to how bad that is since we are working with the same data set and its just a filter applied to that data. Also it means that this store doesn't need to know about each tab that has logic like this and its only going to get worse when we introduce concepts like sorting, etc.

fake/fake-extension.js
avanderhoorn: Good fix.

src/request/component-models/ComponentModel.ts
avanderhoorn: Sounds fine.

src/request/stores/request-detail-store.js
philliphoff: The logging filters are supposed to be sticky, persisting across tabs and selected requests, meaning they outlive the lifetime of the control/model/tab, so this store seems like the right "level" to store that state.  I'm not sure I'd agree that it's "short circuiting" the update chain; the component invokes the component model in response to the user's clicks, the component model transforms the state and publishes the update, the store processes the update and publishes its own update, which is picked up by the detail component, which starts the cascade of re-rendering which will ultimately reflect the filter change.  

(There are lifetime issues with the component model for listening to and relying events because it has no other knowledge about the component lifecycle, for example, with which to properly subscribe/unsubscribe from events and avoid leaks.  I'm a little concerned about introducing those concepts to component models as it feels a little too much like reimplementing parts of Redux, when we might just be better off adopting it itself.)  

`react-redux` has a concept of a "connected component" which is one that has no UI itself but simply manages the properties/states/callbacks of the "presentational" components within it.  Being "connected" to a store, it can dispatch events as well as subscribe to store state (to pick up changes and force re-rendering).  It serves a lot of the same purpose as our component model, but still being a React component means it natively understands the React lifecycle.  We might be able to migrate our idea of a component model to react/redux/react-redux (though the latter requires React 0.14+, to which we have yet to upgrade).

src/request/component-models/LoggingComponentModel.ts
```diff
@@ -1,39 +1,171 @@
...
+
+        if (updated) {
+            this.emitUpdate(state);
+        }
+    }
+
+    public toggleLevel(level: ILoggingLevelModel): void {
+        const state = _.clone(this._requestDetailStore.getState().logging.filter);
+
+        if (state[level.level] === false) {
```
mike-kaufman: `state[level.level] = !state[level.level]`?

src/request/component-models/LoggingComponentModel.ts
```diff
@@ -1,39 +1,171 @@
...
+        return state === undefined || state[level.level] !== false;
+    }
+
+    public toggleAll(): void {
+        const state = _.clone(this._requestDetailStore.getState().logging.filter);
+
+        let updated = false;
+
+        _.forIn(state, (shown, level) => {
+            if (state[level] === false) {
```
mike-kaufman: this isn't accounting for the state[level] === true case.  Is that intentional?  If so, why call this method `toggleAll()`?

src/request/component-models/LoggingComponentModel.ts
philliphoff: `state[level.level]` might === undefined (which I'm treating the same as `true`).

src/request/component-models/LoggingComponentModel.ts
philliphoff: It's intended to "toggle all on".  A better name might be `showAll()`.

test/request/component-models/LoggingComponentModel.spec.ts
```diff
@@ -0,0 +1,534 @@
+'use strict';
```
mike-kaufman: good to see tests.

src/request/component-models/LoggingComponentModel.ts
```diff
@@ -1,39 +1,171 @@
...
+    }
+
+    public isShown(level: ILoggingLevelModel): boolean {
+        const state = this._requestDetailStore.getState().logging.filter;
+
+        return state === undefined || state[level.level] !== false;
+    }
+
+    public toggleAll(): void {
+        const state = _.clone(this._requestDetailStore.getState().logging.filter);
```
mike-kaufman: why are you cloning the object here?  Don't you need to set it back at some point?

src/request/component-models/LoggingComponentModel.ts
mike-kaufman: yeah I like `showAll()`

src/request/component-models/LoggingComponentModel.ts
philliphoff: It's cloned because, in Redux (which our architecture parallels in many ways, and to which we might decide to migrate), state is always replaced rather than mutated.  This new state is published/dispatched and picked up by the store, which will then cause the UI to refresh and reflect the new state.

mike-kaufman:
Would be interesting to see how responsive the UI feels with a large number of log messages.  E.g.,500?

src/request/component-models/LoggingComponentModel.ts
philliphoff: Changed.

philliphoff:
> Would be interesting to see how responsive the UI feels with a large number of log messages. E.g.,500?

I experimented with a request that generated 1500 messages; the UI feels pretty responsive at that level.  I think if we were to adopt Redux, we could probably do better as it makes it pretty easy to ignore re-renders for components not affected by the state change.

mike-kaufman:
Great about the responsiveness.  This LGTM.  What needs to happen to merge this in?

philliphoff:
Just waiting for @avanderhoorn to also sign-off.

src/request/stores/request-detail-store.js
avanderhoorn: Sounds great!

src/shell/components/shell-view.scss
avanderhoorn: After syncing with @chetsangnil none of these lines should change now... He had an internal design review meeting and they said leave it as it was originally.

avanderhoorn:
After that one minor style change/revert, :shipit:

Issue #77, opened 2016-04-29
[Logging] Highlight message tokens


philliphoff:
Add highlighting of the individual tokens within a logging message (if messages were formatted).

PR #85, opened 2016-05-12
Add highlighting of "replaced" log message values

philliphoff:
The Client now adds a subtle highlight to values inserted into a formatted log message (as opposed to being static text).  As shown in the screenshot, inserted values are now **bold**.

screen shot 2016-05-11 at 5 52 52 pm


philliphoff:
Associated with #77.

test/request/component-models/LoggingComponentModel.spec.ts
```diff
@@ -10,6 +10,307 @@ import * as chai from 'chai';
 
 const should = chai.should();
 
+describe('LogMessageModel', () => {
```
mike-kaufman: Love the thorough tests!

mike-kaufman: Probably should add a test case for where spans run back-to-back.  e.g., [0,3], [3, 6].   Probably also want to add a test case for when a span is at the end of the string (e.g., [0,3] for a string of length 4.

mike-kaufman:
LGTM.  Minor comments about some additional test cases.

avanderhoorn:
Once @mike-kaufman stuff it looked at, looks good to me :shipit:

test/request/component-models/LoggingComponentModel.spec.ts
philliphoff: I added tests for back-to-back spans as well as when the message represents one entirely-replaced span.  I already have tests for spans at the very beginning and end of the message, so that should be covered.

Issue #79, opened 2016-04-29
[Logging] Highlight log levels


philliphoff:
Show logging messages using the appropriate color and with the appropriate icon (if any) for their logging level.

philliphoff:
Coloring should be taken care of; the remaining task should just be adding an appropriate icon for warning/error.

PR #87, opened 2016-05-13
Add error/warning icons to log messages

philliphoff:
The logging tab now shows error/warning icons for those respective log messages.

screen shot 2016-05-12 at 4 55 14 pm

The icons are (essentially embedded) [SVGs generated from FontAwesome glyphs](https://github.com/encharm/Font-Awesome-SVG-PNG) (as opposed to using the FontAwesome font directly).


philliphoff:
Associated with #79.

src/shell/components/FontAwesomeIcon.tsx
```diff
@@ -0,0 +1,21 @@
+'use strict';
+
+import React = require('react');
+
+export class FontAwesomeIcon extends React.Component<{ path?: string }, {}> {
```
avanderhoorn: Instead of calling this FontAwesomeIcon, can we call this something like `ShellIcon`? My thinking is that we will be getting out own icons in time and instead of needing to rename post that we can just update the `paths`.

philliphoff: There are some assumptions related to the paths of FontAwesome SVGs (such as a specific viewBox) that might not apply to our own, custom icons and that is why I created a specific component for FontAwesome icons.

philliphoff: I could see this component being a specialization of a more generic SVG component if/when we actually get custom icons as opposed to just borrowing more from FontAwesome.

avanderhoorn:
Synced off line on some of this and its good to go :shipit:

PR #88, opened 2016-05-13
[HttpClient] Component Model and TypeScript conversion

avanderhoorn:
As it says on the box. Still need to get some more tests in and handle the click events on the rows to drive the master/details experience.


avanderhoorn:
I think this can be removed give the direction we have since gone.

PR #93, opened 2016-05-17
Adds formatting for logged objects

philliphoff:
The Logs tab will now offer the user a formatted display of JavaScript objects logged to the console.  Objects will show an expand/collapse icon and, when expanded, use Highlight.js to show a multi-line view of the object with syntax highlighting.  Long messages can also be expanded to see their entirety by hovering over the message with the mouse to display the expand/collapse icon.

screen shot 2016-05-17 at 10 16 57 am

Note: currently all text, truncated or not, can be "expanded".  I'll be looking into a reasonable way to limit that to only text that's actually truncated.


src/request/component-models/LoggingComponentModel.ts
```diff
@@ -44,6 +71,28 @@ export class LogMessageModel implements ILogMessageModel {
...
+        
+        for (let i = value.length - 1; i >= 0 && i >= lastWindowIndex; i--) {
+            if (value[i] === term) {
+                return i;
+            }
+        }
+        
+        return -1;
+    }
+
```
avanderhoorn: Just wondering if we should store the open/closed state on the underlying `_message` object? If we did it means that as someone moves between tabs, that messages would still be open/closed when they move back to the logging tab.

src/lib/modules/glimpse-highlight.js
```diff
@@ -4,5 +4,6 @@ var hljs = require('../../../node_modules/highlight.js/lib/highlight.js');
 
 hljs.registerLanguage('cs', require('../../../node_modules/highlight.js/lib/languages/cs'));
 hljs.registerLanguage('sql', require('../../../node_modules/highlight.js/lib/languages/sql'));
+hljs.registerLanguage('javascript', require('../../../node_modules/highlight.js/lib/languages/javascript'));
```
avanderhoorn: I'm assuming this gives us json.

src/request/component-models/LoggingComponentModel.ts
```diff
@@ -20,6 +22,31 @@ export class LogMessageModel implements ILogMessageModel {
...
     }
 
+    public get isObject(): boolean {
+        
+        //
+        // NOTE: Our heuristic for determining whether text represents an object rather simplistic 
+        //       (to minimize impact on performance). We simply look for starting and ending braces
+        //       (i.e. '{' and '}') near the beginning and ending of the text, respectively, allowing 
+        //       for whitespace and other text that might pre-fix/post-fix the actual object.
+        //
```
avanderhoorn: I think this is reasonable.

src/request/components/request-detail-panel-logging.tsx
```diff
@@ -2,10 +2,71 @@
...
+            isExpanded: !this.state.isExpanded
+        });
+    }
+
+    private getIconClass() {
+        if (this.state.isExpanded) {
+            return 'tab-logs-table-message-icon-expandable';
+        }
+        else if (this.props.message.isObject) {
+            return 'tab-logs-table-message-icon-expandable';
```
avanderhoorn: Should this be wrapped up as an `||` in the above `if` block?

src/request/components/request-detail-panel-logging.tsx
```diff
@@ -2,10 +2,71 @@
...
+        return this.state.isExpanded
+            ? 'tab-logs-table-message-multiline'
+            : '';
+    }
+
+    private getIconPath() {
+        return this.state.isExpanded
+            ? FontAwesomeIcon.paths.CaretDown
+            : FontAwesomeIcon.paths.CaretRight;
+    }
```
avanderhoorn: Maybe `getExpansionIconPath`?

src/request/components/request-detail-panel-logging.tsx
```diff
@@ -2,10 +2,71 @@
...
+        else {
+            return 'tab-logs-table-message-icon';
+        }
+    }
+
+    private getMessageClass() {
+        return this.state.isExpanded
+            ? 'tab-logs-table-message-multiline'
+            : '';
+    }
```
avanderhoorn: Is this only used in the collapsed case?

src/request/components/request-detail-panel-logging.tsx
```diff
@@ -36,15 +97,15 @@ export class Logging extends React.Component<ILoggingProps, {}> {
...
-                                <th width='5%'><span className='table-col-title'>Ordinal</span></th>
-                                <th width='10%'><span className='table-col-title'><FontAwesomeIcon path=''/>Level</span></th>
-                                <th width='40%'><span className='table-col-title'>Message</span></th>
+                                <th width='10%'><span className='table-col-title'>Ordinal</span></th>
+                                <th width='10%'><span className='table-col-title tab-logs-table-icon-column'><FontAwesomeIcon path=''/>Level</span></th>
+                                <th width='58%'><span className='table-col-title tab-logs-table-icon-column'><FontAwesomeIcon path=''/>Message</span></th>
                                 <th width='10%'><span className='table-col-title'>From Start</span></th>
                                 <th width='10%'><span className='table-col-title'>Duration</span></th>
-                                <th />
+                                <th width='2%' />
```
avanderhoorn: What is the column for?

src/request/components/request-detail-panel-logging.tsx
```diff
@@ -2,10 +2,71 @@
...
+
+    public render() {
+        return (
+            <div className='tab-logs-table-message'>
+                <div className={this.getIconClass()} onClick={e => this.onToggleExpansion()}><FontAwesomeIcon path={this.getIconPath()} /></div>
+                {
+                    (this.state.isExpanded && this.props.message.isObject)
+                        ? <div className='tab-logs-table-message-object'><Highlight language='javascript'>{this.props.message.message}</Highlight></div>
+                        : <div className={this.getMessageClass()}>{this.props.message.spans.map(span => <span className={span.wasReplaced ? 'tab-logs-table-message-replaced-region' : ''}>{span.text}</span>)}</div>
+                }
```
avanderhoorn: Does this take into account expanding truncated text?

src/request/components/request-view.scss
```diff
@@ -377,54 +377,96 @@
...
+                
+                > pre {
+                        margin: 0;
+                        
+                        > code {
+                            background: transparent;
+                            padding: 0;
+                        }
+                    }
+            }
```
avanderhoorn: Just be careful about how nested these get. Its easy to forget that these get blown up to rather long selectors when they are compiled. Additionally kinda against best practice to go more than one/two levels deep in cases like this. The fact that they are prefixed with `tab-logs` should be enough specificity, otherwise you end up getting into specificity wars as the app gets bigger.

typings-custom/react-highlight.d.ts
```diff
@@ -1,8 +1,8 @@
 declare module 'react-highlight' {
     import * as __React from 'react';
 
-    interface Highlight extends __React.Component<any, any> {
+    class Highlight extends __React.Component<{ language: string }, {}> {
     }
 
-    export default Highlight;
+    export = Highlight;
```
avanderhoorn: Is this how `default`s are done in TS?

src/lib/modules/glimpse-highlight.js
philliphoff: This actually doesn't appear to be needed; I see the same behavior without it.  It's here just for consistency.  `highlight.js` makes a distinction between JSON and JavaScript in terms of formatting; what we get from `console.log()` is technically not JSON but JavaScript (the only real difference is that the former quotes property names and the latter does not, but it doesn't seem to matter to the formatter anyway).

src/request/components/request-detail-panel-logging.tsx
philliphoff: I've done exactly that in a following commit.

src/request/component-models/LoggingComponentModel.ts
philliphoff: I've also updated the property so that the heuristic is only ever executed once per model.

src/request/components/request-detail-panel-logging.tsx
philliphoff: Yes, the `getMessageClass()` returns a truncated/wrapping style based on the expansion state.

src/request/components/request-view.scss
philliphoff: I was seeing more of this (nested) CSS style when looking through the web and it made it easier for me to understand who "owned" the style.  I do agree that the prefixing in the names are then somewhat redundant.

typings-custom/react-highlight.d.ts
philliphoff: It was the only way TS would stop complaining. :-)  I think this file made its way into the repo as a stray before it was actually used, which is what necessitated the change.

src/request/components/request-detail-panel-logging.tsx
philliphoff: @chetsangnil's design was for the last column in the table to be fixed-width, offset from the end of the table (i.e. not taking the remainder of the available width, which is quite often the style of VS (Windows) tables).  While I'd guess that we might be able to get the same effect via CSS, the simplest way seemed to be just adding a "dummy" column.

src/request/components/request-detail-panel-logging.tsx
philliphoff: Good suggestion; I thought it was a bit generic, too.

src/request/components/request-detail-panel-logging.tsx
philliphoff: No, `getMessageClass()

src/request/component-models/LoggingComponentModel.ts
philliphoff: Agree that maintaining that state is desirable, but (currently) a component model's lifetime is bound to the component's lifetime (the goal of the model to separate UI logic from business logic for test purposes and not so much to persist state long-term), which means that switching tabs may cause both to be re-created.  (This is one reason why the log level filter state is persisted higher-up, so as to be available as the user switches tabs and/or requests.)

We could change the lifetime of the model, of course, but that also means the model comprehend changes in application (i.e. request) state being pushed down from the parent panels and recognize, for example, when the request has changed (e.g. new messages available) in order to choose to regenerate its state (e.g. ordered/filtered messages and level counts).  Going down that route gets us even further in to `redux` territory and its concept of immutable application state which makes it simpler to identify state changes.  At that point, it may just be better to adopt `redux` instead.

src/lib/modules/glimpse-highlight.js
avanderhoorn: Is it worth removing is we don't need it?

src/request/component-models/LoggingComponentModel.ts
avanderhoorn: Agreed, lets hold of on "maintaining" state for when we pickup on redux.

src/request/component-models/LoggingComponentModel.ts
avanderhoorn: Great!

src/request/components/request-detail-panel-logging.tsx
avanderhoorn: >is used to style the plain-text element in both collapsed

Seems a little weird then in terms of logic. They way I read this, is that if the expanded, then apply the multiline class.

philliphoff: The same React component is created regardless of whether the parent is expanded or collapsed; the only difference is the styling attached.  In this case, only the expanded (multi-line) state has any additional styling (from what the inherited values are).  I originally had a "inline" style defined but it ended up being empty, and empty rules generate CSS linting warnings.  I could just name the "inline" style without actually creating one in the CSS, but that could cause confusion if someone assumes there to be one but can't find it.

src/request/components/request-view.scss
avanderhoorn: Have a look at this - https://smacss.com/book/applicability. I don't think its worth changing this now, but before things are through we will need to cleanup our css a bit... note this isn't targeted at this specific code, but I'm guilty of not practicing what I preach here. Hence when we get the new shell design through, we will try and start off on the right foot.

typings-custom/react-highlight.d.ts
avanderhoorn: Sounds fine.

avanderhoorn:
Ok with the exception of the `hljs.registerLanguage` comment, I think this is ready to go :shipit:

src/lib/modules/glimpse-highlight.js
philliphoff: Yes.  I've created refactoring issue #95 for that, as this PR is big enough as it is.

src/request/components/request-view.scss
philliphoff: Thanks, I'll take a look.  I agree that we'll need to do some CSS refactoring as we redesign some of the major Client components and intend to share more common elements.  One of the things that I've wondered about are how to minimize the coupling within the code and the CSS while still allowing for a layered approach to styling.  (For example, I noticed that the log table adds a `table-striped` value to the class name and it seemed a little odd to me to have the code referring to what should be a purely stylistic concept.)

philliphoff:
Associating with #96 as a remaining decision/task to be made.

avanderhoorn:
Do we think we good with just pulling this or should that lang pack be removed if its not really being used?

philliphoff:
I think we're good to merge.  I created #95 to remove the (seemingly) redundant language registration.

Issue #110, opened 2016-05-20
[DataAccess] Display command in split-component


philliphoff:
Display the highlighted and (if possible) formatted data operation command in the command (i.e. left) split-component when the operation is selected.

Issue #109, opened 2016-05-20
[DataAccess] Add command and summary components


philliphoff:
Add the resizable "details" split-component to the data access tab.  Refactor and reuse components from the web services and request tabs (if reasonably possible).

Note: this doesn't include population of the content within those split-components.

philliphoff:
The only remaining item here is making the split-panel resizable.  I experimented with available third-party React components, but they either didn't work as advertised (some had a dependency on there being set widths/heights all the way up the ancestor graph) and some are dependent on later version of React.

I recommend we split re-sizability into a separate issue for tracking and prioritization.

philliphoff:
Split out the resizing-issue so this can be closed.

Issue #105, opened 2016-05-20
[DataAccess] Add data access tab and table


philliphoff:
Add the data access tab and fill out the basic table.  Show Mongo DB related operations as single-line rows, with basic information for each column.

PR #114, opened 2016-05-27
Initial support for the Data tab

philliphoff:
Adds a basic implementation of the Data tab, consisting of a table showing (in particular, Mongo) database operations and its associated metrics.  When an operation is selected, the command (bottom-left) panel will display a formatted version of the operation "command".  (For SQL, this would be the SQL command; for MongoDB, this would be the options JSON.)

screen shot 2016-05-26 at 5 17 22 pm

Notes:
- This PR introduces `redux`, `react-redux`, and `reselect`; the tab is the first to be implemented using that infrastructure.
- At the moment, the `redux` "store" only contains the data needed by this new tab; @avanderhoorn is working on how `redux` may be applied to the Client as a whole.
- At the moment, the `redux` store stores a transformation of the messages that represent data operations.  I've been having an internal debate about whether the store should instead contain the raw messages and with the transformation done instead as a "selector" for the table component.  I think the former is preferable now, given the simplicity of the store and as it won't overlap too much of @avanderhoorn's work.  The good news is that this would be a pretty simple refactoring that has zero impact on the UI components (yay for `redux`!).
- @avanderhoorn might notice that the "master-detail" component is re-implementation of the one used by the Web Services tab, in particular, one based on `flex-box` rather than `table`.  I couldn't get the scrolling/sizing behavior I needed borrowing that tab's implementation so I created one from scratch.  The Web Services tab should be able to move to this, now totally abstract, layout component.


philliphoff:
Associated with #105, #109, #110.

src/request/actions/ActionCreator.ts
```diff
@@ -0,0 +1,32 @@
...
+
+interface IActionCreator<TPayload> extends IActionCreatorBase<TPayload> {
+    (payload: TPayload): IActionCreatorAction<TPayload>;
+}
+
+export function createActionCreator<TPayload>(type: string): IActionCreator<TPayload> {
+    const actionCreator = (payload: TPayload) => {
+        return {
+            type: type,
+            payload: payload
```
avanderhoorn: As a more general question, I wonder if this should be `{ type, ...payload}` instead of having payload hanging off a property?

src/request/components/MasterDetailTriPanel.tsx
```diff
@@ -0,0 +1,34 @@
...
+                    <div className='master-detail-tri-panel-detail-margin' />
+                    <div className='master-detail-tri-panel-detail-panel'>
+                        <div className='master-detail-tri-panel-detail-panel-title'>{this.props.rightDetailPanelTitle}</div>
+                        <div className='master-detail-tri-panel-detail-panel-content'>{this.props.rightDetailPanel}</div>
+                    </div>
+                </div>
+            </div>
+        );
+    }
+}
```
avanderhoorn: Great use of components here.

src/request/components/MasterDetailTriPanel.tsx
```diff
@@ -0,0 +1,34 @@
+import * as React from 'react';
+
+interface IMasterDetailTriPanelProps {
+    masterPanel;
+    leftDetailPanel;
+    leftDetailPanelTitle: string;
+    rightDetailPanel;
+    rightDetailPanelTitle: string;
+}
```
avanderhoorn: Why `Tri`? I would have thought something like `MasterDetailSplitPanel`.

src/request/components/RequestDetailPanelData.tsx
```diff
@@ -0,0 +1,48 @@
...
+        return (
+            <div className='tab-data'>
+            {
+                this.props.totalOperationCount > 0
+                    ? <MasterDetailTriPanel masterPanel={this.renderMaster()} leftDetailPanel={this.renderLeftDetail()} leftDetailPanelTitle='Command' rightDetailPanel={this.renderRightDetail()} rightDetailPanelTitle='Summary' />
+                    : <div className='tab-section text-minor'>Could not find any data.</div>  
+            }
+            </div>
+        );
+    }
```
avanderhoorn: Not for now, but as a more general thing, I wonder if the "empty" case should be handled by the target component? I would mean that all the sites that use it don't have to have this repeated logic.

src/request/actions/ActionCreator.ts
philliphoff: The `redux` documentation recommends [`flux-standard-action`](https://github.com/acdlite/flux-standard-action) as the pattern for actions, which suggests the `payload` property instead of an "inherited" model.  This also tended to simplify "unwrapping" the actions in the reducers from a TypeScript-typed perspective.

src/request/components/RequestDetailPanelData.tsx
```diff
@@ -0,0 +1,48 @@
...
+        )
+    }
+    
+    private renderRightDetail() {
+        return null;
+    }
+    
+    private getTotalOperationCountText() {
+        return this.props.totalOperationCount + (this.props.totalOperationCount === 1 ? ' Operation' : ' Operations');
+    }
```
avanderhoorn: Again more generally, I wonder if its worth having an internal pluralization service that in case like this, you would pass in the word and the count and it would manage the lookup/conversation. This would also more things easier in the future when it comes to internationalization. But all of this is outside the scope of this PR.

src/request/components/MasterDetailTriPanel.tsx
philliphoff: `Tri` was just the first thing I thought of.  To me, "split" would refer to the split between master/detail panels, not that the detail panel itself is further split into two, hence wanting to emphasize the idea that there were three panels.

src/request/components/RequestDetailPanelDataOperationCommand.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+export class RequestDetailPanelDataOperationCommand extends React.Component<IRequestDetailPanelDataOperationCommandProps, {}> {
+    public render() {
+        return (
+            <div className='tab-data-operation-command'>
+                <Highlight language={this.props.language}>{this.props.command}</Highlight>
+            </div>
+        );
+    }
+}
```
avanderhoorn: Do we gain much by this being a separate component instead of just being in `DataComponent.tsx`?

src/request/components/request-view.scss
```diff
@@ -376,6 +376,69 @@
...
+
+.tab-data-operation-table {
+    table-layout: fixed;
+}
+
+.tab-data-operation-table-command-column {
+    overflow: hidden;
+    white-space: nowrap;
+}
+
```
avanderhoorn: Did you test the use case of what it looks like with 100 rows in the master table? Does it scroll correctly, etc?

src/request/components/RequestDetailPanelData.tsx
philliphoff: Yes, I thought the same thing.  In previous teams we had a common "pass through" layout panel that would handle the getting-data/no-data/failed-to-get-data cases in a consistent way.  Definitely something to refactor towards as a separate issue.

src/request/containers/RequestDetailPanelDataContainer.ts
```diff
@@ -0,0 +1,21 @@
...
+
+function mapDispatchToProps(dispatch) {
+    return {
+    };
+}
+
+export = connect(
+    mapStateToProps,
+    mapDispatchToProps
+)(DataComponent);
```
avanderhoorn: Given that `mapDispatchToProps` doesn't return anything, I think `connect` can be updated to just be `connect(mapStateToProps)`. But if you have it like this because you are expecting there to be more actions that come along then thats fine.

src/request/components/RequestDetailPanelData.tsx
philliphoff: Yeah, pluralization tends to be highly-contextual, meaning that automating that in a service is difficult.  (As soon as you get into languages that have gender-ized articles, like German, the entire  phrase has to be localized, not just individual words.)  You pretty much end up with the same code that's here--a test for 0/1/>1 and then the resource strings that represent each case.  You might be able to abstract away the test, but you still have to pass the resource string names for each case.

src/request/containers/RequestDetailPanelDataOperationCommandContainer.ts
```diff
@@ -0,0 +1,33 @@
...
+
+function mapDispatchToProps(dispatch) {
+    return {
+    };
+}
+
+export const RequestDetailPanelDataOperationCommandContainer = connect(
+    mapStateToProps,
+    mapDispatchToProps
+)(RequestDetailPanelDataOperationCommand);
```
avanderhoorn: Same comment here as with `RequestDetailPanelDataContainer`

src/request/components/RequestDetailPanelDataOperationCommand.tsx
philliphoff: The "parent" component is a "dumb" component (in the `redux` sense); it knows where to place the component (in the left master/detail/split panel) but doesn't know how to pass it the correct set of properties (because those are determined by application state and set by the table component).

src/request/containers/RequestDetailPanelDataOperationCommandContainer.ts
avanderhoorn: Also wondering if this `mapStateToProps` should happen here, instead of within `RequestDetailPanelDataContainer`? Given that `RequestDetailPanelDataOperationCommandContainer` is now "smart" if this is ideal or not. I'm assuming you have seen https://medium.com/@dan_abramov/smart-and-dumb-components-7ca2f9a7c7d0#.xvswxbfmp.

src/request/components/request-view.scss
philliphoff: Yes, the "master" panel will scroll in that case. (There is a case where, if the browser window is too small, the Client's request details panel will itself also scroll, which is a little weird, but I believe it's due to the styling of the details panel itself.  The Web Services tab behaves the same way.)

src/request/containers/RequestDetailPanelDataOperationTableContainer.ts
```diff
@@ -0,0 +1,28 @@
...
+        onSelected: (index: number) => {
+            dispatch(selectOperationAction(index));           
+        }
+    };
+}
+
+export const RequestDetailPanelDataOperationTableContainer = connect(
+    mapStateToProps,
+    mapDispatchToProps
+)(RequestDetailPanelDataOperationTable);
```
avanderhoorn: Same questions here as the above smart vs dumb.

src/request/containers/RequestDetailPanelDataContainer.ts
philliphoff: The [`react-redux` documentation](https://github.com/reactjs/react-redux/blob/master/docs/api.md#connectmapstatetoprops-mapdispatchtoprops-mergeprops-options) states that, if `mapDispatchToProps` is omitted, `connect` will then map `dispatch` into the wrapped component's properties.  While that probably wouldn't hurt, it's not exactly what I would have expected, so I explicitly pass a "dummy" implementation.

src/request/actions/ActionCreator.ts
avanderhoorn: Great! Hadn't picked up on that.

src/request/components/MasterDetailTriPanel.tsx
avanderhoorn: Ok I can get behind that.

src/request/components/RequestDetailPanelData.tsx
avanderhoorn: Great.

src/request/components/RequestDetailPanelData.tsx
avanderhoorn: My main thought was something simple for now. I know it gets fairly complex fairly quickly, but the alternative is that we start getting code like this all over the place. Hence just wondering if there is something simple we can do in the mean time.

src/request/components/RequestDetailPanelDataOperationCommand.tsx
avanderhoorn: Ok that makes sense.

src/request/components/request-view.scss
avanderhoorn: Ya, hopefully that case will be picked up when we do the larger shell conversion.

src/request/containers/RequestDetailPanelDataContainer.ts
avanderhoorn: Arr ok, makes sense. Does passing null in work (I'm guessing that it would cause the same problem)?

philliphoff: Not sure; the documentation doesn't specify.  It looks like I could simply pass `{}` however (which `react-redux` treats as an object of action creators and where it would wrap each property in a dispatch-passed function and merge with the component's properties, but since there are none, essentially nothing would happen).  It's less code, which would is generally good, but it's also less explicit (that there are no dispatch properties), which can sometimes hinder understanding.

src/request/containers/RequestDetailPanelDataOperationCommandContainer.ts
philliphoff: Yes, did read the post (and similar others).  There seems to be a fairly liberal application of the smart vs. dumb component idea--liberal in the sense of a nearly 1:1 mapping between them as opposed to trying to minimize the number of smart components.  The posts suggest that the need for dumb components to pass down properties to their child components, without using them itself, is often a sign that there should be a smart component in between.

As the Data tab evolves it's going to have more and more sub-components (as each sub-tab of the detail panel are added), which would compound the problem of passing down the set of properties necessary for each subcomponent.  If each of those has their own, corresponding smart component, then it becomes very obvious what is responsible for getting/setting their properties.

src/request/actions/ActionCreator.ts
philliphoff: Also note that this `ActionCreator` is a bit of an experiment to try to elimination of duplicated strings (i.e. the action types) between smart component and reducer, while at the same time coupling action types to the data associated with them.  (The former would often be done through string constants, but I've yet to see anybody tackle the latter.)

src/request/containers/RequestDetailPanelDataContainer.ts
avanderhoorn: Understood. Lets leave it for now and as more patterns evolve we can adjust then.

src/request/containers/RequestDetailPanelDataOperationCommandContainer.ts
avanderhoorn: I can see that logic. I hadn't interpreted it that way, but reading through it again I can see that.

src/request/actions/ActionCreator.ts
avanderhoorn: I noticed that and I think the general pattern probably also helps with TS typing, etc.

src/request/reducers/RequestDetailDataReducer.ts
```diff
@@ -0,0 +1,174 @@
...
+interface ISortableOperation {
+    ordinal: number;
+    operation: IRequestDetailDataOperationState;
+}
+
+function updateSelectedIndex(state: number, request) {
+    return request
+        ? state
+        : 0; 
+}
```
avanderhoorn: nit, any reason this is over 3 lines than just one?

src/request/actions/ActionCreator.ts
```diff
@@ -0,0 +1,32 @@
...
+        }
+    };
+    
+    const typedActionCreator = <IActionCreator<TPayload>>actionCreator;
+    
+    typedActionCreator.type = type;
+    typedActionCreator.unwrap = (action: Action) => {
+        return (<IActionCreatorAction<TPayload>>action).payload;
+    };
```
avanderhoorn: I'm guessing the reason to have `unwrap` as apposed to just getting the payload property is because this method allows us to access the payload already cast?

philliphoff: Yes, the `Action` comes into the reducer untyped; it's only after the switch do you know the expected type.  You could arbitrarily cast it to the correct type, but now there's a disconnect between the action type and interface.  The `unwrap()` function is an attempt to couple the two together.  (It also makes the reducer easier to read by eliminating the TypeScript casting, which I find hard to read.)

avanderhoorn: I think this is a reasonable approach to the problem. We might even be able to abstract this out later on as a more generalized comment for people creating actions using typescript.

avanderhoorn:
I'm good for this to go in :shipit:

src/request/components/RequestDetailPanelData.tsx
philliphoff: What this would probably look like is:

```javascript
function pluralize(testValue: number, oneFormatString: string, manyFormatString: string, zeroFormatString: string, formatArgs?: string[]);
```
Example:

```javascript
const oneOperation = '%n Operation';
const zeroOrManyOperations = '%n Operations';
<div>{pluralize(count, oneOperation, zeroOrManyOperations, zeroOrManyOperations, [ count ])}</div>
```
You pass in the test value, then the (optionally formatted) resource strings for each case, then (optionally) the argument list for any formatted strings.  If no argument list was passed, then maybe you could assume that the test value is the implied first argument (if the strings were formatted).  That would allow you to drop the args list in most simple cases.

src/request/reducers/RequestDetailDataReducer.ts
philliphoff: No just habit, as I find it easier to read the test/true/false cases on separate lines, in particular, when any of them are longer statements (such as HTML elements in TSX files).

PR #117, opened 2016-05-31
Integrate Redux DevTools Chrome Extension Client

philliphoff:
Integrates the [Redux DevTools Chrome Extension](https://github.com/zalmoxisus/redux-devtools-extension) with the Client (if installed).  The extension allows you to see `redux` actions and as they're dispatched and the resulting changes to application state as well as  "time travel" by rolling back/forward actions.

screen shot 2016-05-31 at 3 32 06 pm

Note: this is the "lightweight" version integration that only works with the Chrome extension, but it prevents us from having to embed the tools within the Client itself.

### DIFF
--- a/src/request/stores/RequestStore.ts
+++ b/src/request/stores/RequestStore.ts
@@ -2,4 +2,5 @@ import { requestReducer } from '../reducers/RequestReducer';
 
 import { createStore } from 'redux';
 
-export = createStore(requestReducer);
+export = createStore(requestReducer,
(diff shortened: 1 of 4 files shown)


avanderhoorn:
I'm wondering if we want to put a switch in here so that the tools are explicitly excluded if we are in non debug mode...

philliphoff:
That's a good idea.  I've updated the PR with a Webpack switch to enable/disable the development tools (defaulting to disabled).

Issue #106, opened 2016-05-20
[DataAccess] Add filtering of databases


philliphoff:
Add a filter bar to the data access tab that allows the user to show/hide operations related to certain types of databases (e.g. MongoDB, MySQL).  

Questions: 
- [x] Is the set of filters well known, or derived based on the messages published during the request?
- [x] If messages are found only pertain to a single database type, should the filter bar be shown?
- [x] Is there a preference for display order of database types (e.g. arbitrary, alphabetical, by count, etc.)

This likely implies some refactoring of the filter bar for the logging tab in order to create a single component/style that can be re-used between them and other tables in the future.

PR #118, opened 2016-06-01
Add filtering to the Data tab

philliphoff:
Adds filtering to the Data tab of the Client.  As requests are selected, the Client will track what database-related messages have been observed across requests and show those databases in a "filter bar" above the operations table, along with the number of operations pertaining to each for that selected request.  The user can then show/hide operations associated with each database.

screen shot 2016-06-01 at 1 18 42 pm


philliphoff:
Associated with #106.

src/request/actions/ActionCreator.ts
```diff
@@ -30,3 +34,17 @@ export function createActionCreator<TPayload>(type: string): IActionCreator<TPay
...
+            type: type        
+        }
+    };
+    
+    const typedActionCreator = <ISimpleActionCreator>actionCreator;
+    
+    typedActionCreator.type = type;
+    
+    return typedActionCreator;
+}
```
avanderhoorn: Is this just an action without a payload?

philliphoff: Yes.  We could have used a singleton action instance, but for the sake of consistency with payload-bearing actions it made sense to keep the same action-factory pattern.

src/request/components/FilterBar.tsx
```diff
@@ -0,0 +1,34 @@
...
+            </div>
+        );
+    }    
+    
+    private renderFilter(filter: IFilterButtonProps) {
+        return (
+            <FilterButton count={filter.count} isShown={filter.isShown} name={filter.name} onToggle={() => this.props.onToggle(filter.name)} />
+        );
+    }
+}
```
avanderhoorn: Great use of components!

src/request/selectors/RequestDetailDataSelectors.ts
```diff
@@ -1,19 +1,73 @@
...
+            if (!_.includes(hiddenDatabases, operation.database)) {
+                filteredOperations.push({
+                    ordinal: index + 1,
+                    operation: operation
+                })
+            }
+        });
+        
+        return filteredOperations;
+    });
```
avanderhoorn: I'm assuming this is deciding what to include and not include based on its index? If so, longer term when we start introducing things like sorting, additional filtering, etc if we want to base this on a fixed id or something.

src/request/selectors/RequestDetailDataSelectors.ts
```diff
@@ -1,19 +1,73 @@
...
     getOperations,
     operations => {
         return operations.length;
     });
 
+export const getFilteredOperations = createSelector(
+    getFilterState,
+    getOperations,
+    (filterState, operations) => {
+        const hiddenDatabases = _.keys(_.omitBy(filterState, isShown => isShown));
```
avanderhoorn: nit: I wonder... instead of converting this to an array of keys, if its left as an object, then on like 25 we would have to scan through the array, it would just be an index look up.

src/request/selectors/RequestDetailDataSelectors.ts
```diff
@@ -1,19 +1,73 @@
...
+            : selectedOperationIdState;
+    });
+
 export const getSelectedOperation = createSelector(
+    getFilteredOperations,
+    getSelectedOperationId,
+    (operations, selectedOperationId) => {
+        // TODO: Can this (need this) be optimized by building a map of id --> operation?
+        return _.find(operations, operation => operation.operation.id === selectedOperationId);
+    });
```
avanderhoorn: Not sure if it helps, but this is something that I have done with the stuff that's coming that might be useful here - https://github.com/Glimpse/Glimpse.Client.Strawman/blob/master/src/client/modules/messages/reducers/messages-reducer.js#L17-L74.

src/request/selectors/RequestDetailDataSelectors.ts
```diff
@@ -1,19 +1,73 @@
...
+                    (filter, database) => {
+                        const databaseOperations = databases[database];
+                        
+                        return {
+                            name: database,
+                            isShown: filter,
+                            count: databaseOperations ? databaseOperations.length : 0
+                        }
+                    })),
+            filter => filter.name);
```
avanderhoorn: Can this be done better by chaining the operations? At least it might be a little easer to read.

avanderhoorn:
Ok finally done. I've made a few comments, nothing too major just some observations.

src/request/selectors/RequestDetailDataSelectors.ts
philliphoff: Yes, since we don't support arbitrary sorting in the UI that's taken care of by the reducer (to keep things simple and not have to introduce an intermediate selector to do the sorting), which allows us to just iterate over the state and include/exclude based on the filters.  If we changed the UI, for example, to allow an ascending or descending sort, then we would probably shift responsibility for sorting to a selector.

src/request/selectors/RequestDetailDataSelectors.ts
philliphoff: Yep, I think that structure was a holdover from when filters originally was an array rather than a map.  I've made the change.

src/request/selectors/RequestDetailDataSelectors.ts
philliphoff: Yes, I was trying to decide whether it was worth the extra complexity in the reducer of generating both a map and sorted array of operations, just for this lookup, or whether it was simpler and performant-enough, given the likely number of data operations per request, to just scan for the selected ID.  I'm still leaning towards the latter, whereas, in the case of logging, for example, I would probably go with the two mappings (as one would likely expect far more logging statements per request than data operations).

src/request/selectors/RequestDetailDataSelectors.ts
philliphoff: Yes, that went through a number of iterations.  Using chaining now that it's more fixed cleans things up a bit.

src/request/selectors/RequestDetailDataSelectors.ts
philliphoff: I actually did a little spike to see whether two mappings helped at all in the case of the Data tab.  To make it work, we'd need to resort the messages after any filter is applied in order to keep a mapping of filtered IDs to items, as we want ensure that only filtered items can be selected.  So it's a choice of a scan per render or a resort per filter change.  Either approach is probably negligibly different, given the total number of operations per typical request.

PR #120, opened 2016-06-07
Adds sub-tabs to the Data Summary panel

philliphoff:
Adds a basic implementation of the Summary panel sub-tabs for the Data tab. Only the "General Info" sub-tab is implemented (showing basic information for MongoDB operations, as the SQL proxy/messages do not yet capture this information).

screen shot 2016-06-07 at 9 16 30 am


avanderhoorn:
I'm assuming there is a little bit of styling still outstanding? Specifically the font color of the selected tab and the background color behind the tabs (#fbfbfb)?

src/request/components/RequestDetailPanelDataOperationSummary.tsx
```diff
@@ -0,0 +1,54 @@
...
+                    <TabPanel header='Parameters'>
+                    {
+                        this.renderParameters()   
+                    }
+                    </TabPanel>
+                    <TabPanel header='Connection'>
+                    {
+                        this.renderConnection()
+                    }
+                    </TabPanel>
```
avanderhoorn: Does this lazy render the content... I'm guessing so but just wanted to make sure.

src/request/components/RequestDetailPanelDataOperationSummaryGeneralInfo.tsx
```diff
@@ -0,0 +1,37 @@
...
+                <br />
+                <br />
+                <span className='label'>Database: </span><span>{this.props.databaseType}</span>
+                <br />
+                <br />
+                <span className='label'>Database Name: </span><span>{this.props.databaseName || '-'}</span>
+                <br />
+                <br />
+                <span className='label'>Server: </span><span>{this.props.serverName || '-'}</span>
+            </div>
```
avanderhoorn: Ultimately its probably better to do this as a `<ul ...></ul>`. Assuming that any "list" like this implements the same style, we can change the style/layout of all of them in one place. Using `br`s like this make it harder and isn't typically used in cases like this.

src/request/components/TabPanel.tsx
```diff
@@ -0,0 +1,7 @@
+import * as React from 'react';
+
+export class TabPanel extends React.Component<{ header: string, children? }, {}> {
+    public render() {
+        return this.props.children;
+    }
```
avanderhoorn: Is `header` being used for anything?

src/request/components/TabbedPanel.tsx
```diff
@@ -0,0 +1,64 @@
...
+    
+    public render() {
+        return (
+            <div className='tabbed-panel'>
+            {
+                this.renderHeaders()
+            }
+            {
+                this.renderSelectedTab()
+            }
```
avanderhoorn: nit, this could probably be shortened to something like this:
```
            {this.renderHeaders()}
            {this.renderSelectedTab()} 
```

src/request/components/RequestDetailPanelDataOperationSummaryGeneralInfo.tsx
philliphoff: Yeah, I started down the `<ul>` path but saw that it would likely take a lot of restyling (e.g. to get rid of the bullet points), so I chose this as the simpler route.  I can experiment with it again.

src/request/components/RequestDetailPanelDataOperationSummary.tsx
```diff
@@ -0,0 +1,54 @@
...
+                    <TabPanel header='Parameters'>
+                    {
+                        this.renderParameters()   
+                    }
+                    </TabPanel>
+                    <TabPanel header='Connection'>
+                    {
+                        this.renderConnection()
+                    }
```
avanderhoorn: nit, same styling comment as below with the above being:
```
                    <TabPanel header='Parameters'>
                         {this.renderParameters()}
                    </TabPanel>
                    <TabPanel header='Connection'>
                        {this.renderConnection()}
                    </TabPanel>
```

philliphoff:
I'll double-check the sub-tab styling; there was a reason I used that specific selected/unselected foreground color, but I probably just missed the different background color of the headers (as it's not very different in the PDF).

src/request/components/RequestDetailPanelDataOperationSummaryGeneralInfo.tsx
avanderhoorn: Great. If you need any help with the css, just let me know.

src/request/components/TabbedPanel.tsx
```diff
@@ -0,0 +1,64 @@
...
+            <div className={classNames('tabbed-panel-header', { 'tabbed-panel-header-selected': index === this.state.selectedIndex })} key={header} onClick={e => this.onSelect(index)}>{header}</div>
+        );
+    }
+    
+    private renderSelectedTab() {
+        return (
+            <div className='tabbed-panel-selected'>
+            {
+                this.props.children[this.state.selectedIndex]
+            }
```
avanderhoorn: nit, same style as previously mentioned

src/request/components/TabbedPanel.tsx
```diff
@@ -0,0 +1,64 @@
...
+    public select(index: number) {
+        this.onSelect(index);
+    }
+    
+    private renderHeaders() {
+        return (
+          <div className='tabbed-panel-headers'>
+          {
+              this.props.children.map((tabPanel, index) => this.renderHeader(tabPanel.props.header, index))
+          }
```
avanderhoorn: nit, same style as previously mentioned

src/request/components/TabbedPanel.tsx
philliphoff: Probably, but that's so much harder to read. :-)

src/request/components/request-view.scss
```diff
@@ -421,6 +421,43 @@
...
+    color: #7c7c7c;
+    cursor: pointer;
+    margin-bottom: -1px;
+    padding: 5px;
+}
+
+.tabbed-panel-header-selected {
+    @extend .tabbed-panel-header;
+    
+    border-color: $tabbed-panel-border-color $tabbed-panel-border-color $tabbed-panel-background-color $tabbed-panel-border-color;
```
avanderhoorn: Love the use of the vars...

src/request/reducers/RequestDetailDataReducer.ts
```diff
@@ -134,6 +142,8 @@ function createMongoDbDeleteOperation(message: IMessageEnvelope<IDataMongoDbDele
     return {
         id: message.id,
         database: 'MongoDB',
+        databaseName: message.payload.database,
+        serverName: message.payload.connectionHost,
```
avanderhoorn: Wonder if its worth creating a `payload` const and stashing the data there... might save a few bytes. That said, when the product pack is done, it might optimize for cases like this.

avanderhoorn:
All looks good here. Couple of minor styling things, but no major blockers.

src/request/components/TabbedPanel.tsx
avanderhoorn: I get that, but its fairly standard as far as jsx is concerned.

nebrius: The shorthand is how I've always seen it done in React, fwiw

src/request/components/RequestDetailPanelDataOperationSummary.tsx
philliphoff: The `TabbedPanel` will only render the selected `TabPanel` instance, so while the components in each `TabPanel` will have been *created*, they shouldn't actually all be *rendered*.

src/request/components/TabPanel.tsx
philliphoff: Yes, `header` is used by the parent `TabbedPanel` to render the headers of all of the tabs.

src/request/components/TabbedPanel.tsx
philliphoff: It's funny--I was always taught that ample whitespace was a virtue.

philliphoff:
@chetsangnil and I sat down and went over all of the styling.  We should now be pretty close to his original design.

philliphoff:
Associated with #111.

PR #121, opened 2016-06-08
Ported Web Services tab to Redux-based implementation

philliphoff:
This is a straight port of the existing Web Services tab (pure JSX) to a Redux-based implementation.  The only alteration was substitution of its master/detail "tri panel" with the generic one used by the Data tab.

Note: no unit tests were created for this PR, but they will come along on subsequent PRs which actually add new features.


src/request/components/RequestDetailPanelWebServices.tsx
```diff
@@ -0,0 +1,130 @@
...
+                <tbody>
+                    {requestItems}
+                </tbody>
+            </table>
+        );
+    }
+}
+
+/**
+ * React class to display service details
```
mike-kaufman: service details?

mike-kaufman: isn't this "service headers" or "web service headers"?  (I guess I'm confused on where the "details" bit of the comes from)

mike-kaufman:
lgtm, but I haven't been following the client UI stuff closely so @avanderhoorn should sign off.

src/request/components/RequestDetailPanelWebServices.tsx
philliphoff: I didn't do any renaming of the components; at the moment those panels show only the request/response headers but more "details" will eventually shown (e.g. body, cookies, etc.).

src/request/request-tab.js
```diff
@@ -35,9 +35,16 @@ module.exports = {
 
 // TODO: Need to come up with a better self registration process
 require('./components/request-detail-panel-execution');
-require('./components/request-detail-panel-webservices');
```
avanderhoorn: Can this file be removed now?

philliphoff: Yep, it *was* but was brought back to life due to a merge conflict.  It should be dead again.

PR #122, opened 2016-06-08
Ports the Request tab to a Redux-based implementation

philliphoff:
A straight port of the existing Request tab from its pure JSX-based implementation to a `redux`-based one.  No other changes made.

Note: no unit tests were added related to this port; they will be added as new features are developed.


src/request/request-tab.js
```diff
@@ -45,12 +45,12 @@ module.exports.registerTab({
     component: WebServices
 });
 
-var Request = require('./components/request-detail-panel-request').Request;
```
avanderhoorn: Don't think that this file has been removed. Must be same problem as before?

philliphoff: I don't see it in the branch; GitHub is showing the change as a rename/edit (as expected).

avanderhoorn: arr ok sorry about that.

PR #124, opened 2016-06-09
Display captured HTTP server request/response bodies

philliphoff:
A quick refactoring of the Request tab to show any captured request/response bodies as well as to better conform to the latest UX (two columns with sub-tabs).

screen shot 2016-06-09 at 4 43 10 pm

### DIFF
--- a/src/lib/StringUtilities.ts
+++ b/src/lib/StringUtilities.ts
@@ -1,3 +1,6 @@
+/**
+ * Converts strings of the form 'one-two-three' to 'One-Two-Three', also called 'train-case'.
+ */
 export function trainCase(value: string): string {
     if (value) {
         let newValue = '';
--- a/src/lib/StringUtilities.ts
+++ b/src/lib/StringUtilities.ts
@@ -1,0 +1,15 @@
+export function trainCase(value: string): string {
+    if (value) {
+        let newValue = '';
+
+        for(let i = 0; i < value.length; i++) {
+            newValue += (i === 0 || value[i - 1] === '-') 
+                ? value[i].toUpperCase()
+                : value[i];
+        }
+
+        return newValue;
+    }
+
+    return value;
+}
--- a/src/request/reducers/RequestDetailRequestReducer.ts
+++ b/src/request/reducers/RequestDetailRequestReducer.ts
@@ -8,8 +8,14 @@ import { Action } from 'redux';
 
 const defaultState: IRequestDetailRequestState = {
     url: '',
-    requestHeaders: {},
-    responseHeaders: {}
+    request: {
+        body: '',
+        headers: {}
+    },
+    response: {
+        body: '',
+        headers: {}
+    }
 };
 
 // TODO: Consolidate this function into a utility function across reducers.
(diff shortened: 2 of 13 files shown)


src/lib/StringUtilities.ts
```diff
@@ -0,0 +1,18 @@
...
+            newValue += (i === 0 || value[i - 1] === '-') 
+                ? value[i].toUpperCase()
+                : value[i];
+        }
+
+        return newValue;
+    }
+
+    return value;
+}
```
avanderhoorn: Will this work instead? https://lodash.com/docs#kebabCase I know that we are trying to lesson our dependency on lodash, but we can also setup the build to only bring in what we need from lodash.

avanderhoorn:
Looks good besides the small comment I made. Also, I'd like to make sure that the data generator is kept up to date with spitting out data that works for the features we are adding... in this case req/res bodies. Because we are in the process of converting over, I'm no too worried at this point, but keep it in mind for the future.

test/lib/StringUtilities.spec.ts
```diff
@@ -0,0 +1,71 @@
+import { trainCase } from '../../src/lib/StringUtilities';
+
+import * as chai from 'chai';
+
+const should = chai.should();
+
+describe('StringUtilities', () => {
+    describe('#trainCase', () => {
+        it('should return undefined for an undefined value', () => {
```
mike-kaufman: probably should add some cases for strings with non-alpha chars.  like `'1', '1-1'` and just `'-'`.

src/lib/StringUtilities.ts
philliphoff: No, "kebab-case" specifically converts to lower-case.  "Train-case" is similar but uses capitalized words.  `lodash` has a number of these types of conversion functions but not that specific one.

mike-kaufman:
lgtm.

test/lib/StringUtilities.spec.ts
philliphoff: Good ideas; I've added such tests.

src/request/reducers/RequestDetailRequestReducer.ts
```diff
@@ -30,8 +36,14 @@ function updateRequestState(state: IRequestDetailRequestState, request): IReques
...
-            requestHeaders: requestMessage ? requestMessage.payload.headers : undefined,
-            responseHeaders: responseMessage ? responseMessage.payload.headers : undefined
+            request: {
+                body: requestMessage && requestMessage.payload.body && requestMessage.payload.body.content ? requestMessage.payload.body.content : '',
+                headers: requestMessage ? requestMessage.payload.headers : undefined
+            },
+            response: {
+                body: responseMessage && responseMessage.payload.body && responseMessage.payload.body.content ? responseMessage.payload.body.content : '',
+                headers: responseMessage ? responseMessage.payload.headers : undefined
+            }
```
avanderhoorn: Wish we didn't have to do the null checks like this.

src/lib/StringUtilities.ts
avanderhoorn: OK no worries.

PR #138, opened 2016-06-17
Adding basic telemetry instrumentation for glimpse client

mike-kaufman:
- added a TelemetyClient class that will hook up to various shell events and send corresponding App Insights telemetry events. 
 - Pull the telemetry configuration data from a "telemetry-config" resource. 
 - Added "fake data" to support dev-time scenarios.


src/lib/TelemetryClient.ts
```diff
@@ -0,0 +1,166 @@
...
+    constructor() {
+        this.sessionId = uuid.v4();
+        const self = this;
+        metadataRepository.registerListener((metadata) => {
+
+            if (!metadata.resources || !metadata.resources['telemetry-config']) {
+                self.isTelemetryEnabled = false;
+            }
+            else {
+                const uri = metadata.resources['telemetry-config'];
```
avanderhoorn: Now that i think about it, I wonder if to save having to make a second request for this data, we could include this data in the metadata payload...

mike-kaufman: It's only one extra request, so it isn't like it's repeated.  That said, I don't have strong opinions about it, other than what I have is implemented & seems to be working.

mike-kaufman:
ping... I"d like to get this in.

src/lib/TelemetryClient.ts
avanderhoorn: Thats fine for now

PR #140, opened 2016-06-23
Adds middleware execution to Request tab

philliphoff:
Adds a table displaying the execution of middleware for a given request to the Request tab.  The table shows nested middleware (e.g. router pipelines) as well as which piece of middleware set which outgoing header.

screen shot 2016-06-23 at 3 26 35 pm


philliphoff:
Associated with #125, #126.

chetsangnil:
Can we add abit more indentation for the nested items? For example, in ordinal 10, we can add enough indentation so that '<' line up with 't' from "router" above it.

src/request/selectors/RequestDetailRequestSelectors.ts
```diff
@@ -1,3 +1,47 @@
...
-export const getRequest = (state: IRequestState) => state.detail.request;
+import { createSelector } from 'reselect';
+
+const getMiddlewareState = (state: IRequestState) => state.detail.request.middleware;
+
+export const getUrl = (state: IRequestState) => state.detail.request.url;
+export const getRequest = (state: IRequestState) => state.detail.request.request;
+export const getResponse = (state: IRequestState) => state.detail.request.response;
+
+interface IFlattenedMiddleware {
```
avanderhoorn: Do we want to come up with a general name for these "first class" "view models"... like `IMiddlewareViewModel`? I don't think that this needs to be done now and can wait till we get over all the tabs into the strawman. I'm sure that at this point we will see other patterns we want to bring across as well.

src/request/components/RequestDetailPanelRequest.tsx
```diff
@@ -79,4 +90,64 @@ export class Request extends React.Component<IRequestProps, {}> {
             </div>
         );
     }
+
+    private renderMiddleware() {
+        return (
+            <div className='tab-request-middleware'>
```
avanderhoorn: nit, shouldn't these be double quotes?

src/request/reducers/RequestDetailRequestReducer.ts
```diff
@@ -19,42 +26,156 @@ const defaultState: IRequestDetailRequestState = {
...
+                return reducer(
+                    state, 
+                    requestMessage ? requestMessage.payload : undefined,
+                    responseMessage ? responseMessage.payload : undefined);
+            }
+        }
+
+        return state;
+    }
+}
```
avanderhoorn: Whats the intent of this? I'm not quite following. It almost looks like it should be self invoking to work as intended but I'm guessing that this does work.

test/request/reducers/RequestDetailRequestReducers.spec.ts
```diff
@@ -0,0 +1,190 @@
+import { IMessage, IMessageEnvelope } from '../../../src/request/messages/IMessageEnvelope';
+import { IMiddlewareStartPayload } from '../../../src/request/messages/IMiddlewareStartPayload';
+import { IMiddlewareEndPayload } from '../../../src/request/messages/IMiddlewareEndPayload';
```
avanderhoorn: Thinking out loud, regarding conventions around "models" of various types, it would be nice if all models coming from the server had a consistent postfix and models that are constructed for the views have another post fix. Would make it really easy to spot what levels of data are being passed around where.

avanderhoorn:
Looks good. I made a few comments around various items, but the only thing blocking was a nit on some quote usage.

src/request/components/RequestDetailPanelRequest.tsx
philliphoff: No, TypeScript leans towards single quotes (`'`) as string delimiters, and we have linting rules to enforce that (at least, we do in other repos).

src/request/reducers/RequestDetailRequestReducer.ts
philliphoff: To accomodate the addition of a middleware reducer (in particular, to allow testing of the middleware reducer apart from the other request-related reducing), I needed to split apart that "reduction" into each constituent parts (i.e. url, request, and response properties) so they could then be combined with middleware.  Each of those reductions had the same form (i.e. responded to the same actions) so I created a factory (`createRequestReducer`) that generates the boilerplate for handling those actions.

philliphoff: This will be simplified once moved to the strawman Client, as this logic will be moved to a selector and each constituent part will therefore be a simple projection rather than a full-on reducer.

test/request/reducers/RequestDetailRequestReducers.spec.ts
philliphoff: In general, the only models coming from the server are messages and I think we have a fairly straightforward naming convention there (e.g. `IMessage`, `IMessageEnvelope`, `IXxxPayload`, etc.).  I've tried to name "view models" consistently as well (e.g. `IWebServicesProps`, `IWebServicesDispatchProps`, etc.) and in a way that maps well to their containers.

There is the transfer of data between selector and container that could be explicitly described in interfaces, but I've found that it generally doesn't have to as TypeScript does a pretty good job dealing with the implied types.

philliphoff:
> Can we add abit more indentation for the nested items? For example, in ordinal 10, we can add enough indentation so that '<' line up with 't' from "router" above it.

I've added a bit more indentation; it won't be exact though as we're using proportional (i.e. non-mono-spaced) fonts which will affect the apparent indentation based on the naming of the two middleware.

Issue #135, opened 2016-06-10
[HttpServer] Add Raw request/response sub-tab


philliphoff:

philliphoff:
There was some confusion as to what "raw" meant in terms of displaying just the un-highlighted, un-formatted body content or a complete Fiddler-style request/response stream.  The outstanding PR is to implement the former and there are some complications to implementing the latter.

I'm not sure yet where we're landing with this issue.

PR #141, opened 2016-06-24
Add "raw" body request/response sub-tabs to Request tab.

philliphoff:
Adds a "raw" sub-tab to the request/response tabs within the Request tab, displaying the un-formatted, un-highlighted body content of the request/response.

screen shot 2016-06-24 at 9 23 51 am


philliphoff:
Associated with #135.

src/request/components/RequestDetailPanelRequest.tsx
```diff
@@ -79,4 +82,12 @@ export class Request extends React.Component<IRequestProps, {}> {
             </div>
         );
     }
+
+    private renderRaw(body: string) {
+        return (
+            <div className='tab-request-raw'>
```
avanderhoorn: nit, quote usage.

avanderhoorn:
Looks great, just the quote usage nit.

src/request/components/RequestDetailPanelRequest.tsx
philliphoff: TypeScript leans towards single quotes `'` to delimit strings and we have linting rules to that effect (at least in other repos).  Unless, you mean something entirely different.

avanderhoorn:
I just realized, this quite the right functionality for the tab. This tab is supposed to show the "whole" raw req/res. Currently this just shows the raw body, but the whole raw req/res also includes the heads, body, etc.

nikmd23:
Yeah, at @avanderhoorn just mentioned, this should show the raw HTTP request or response.

avanderhoorn:
Are you happy that this is a reconstruction vs an exact copy? An exact copy would require us to essentially have all the data duplicated.

nikmd23:
I'm happy with it being a reconstruction. Just watch things like the sort order for headers, and capitalization, we'll want to get those right.

avanderhoorn:
Re sort order... I'm not 100% sure if we get in at the level where sort order is maintained. Additionally on the response we may not get all headers since some headers are added downstream of where we tap.

philliphoff:
> I'm happy with it being a reconstruction. Just watch things like the sort order for headers, and capitalization, we'll want to get those right.

We won't necessarily get the actual sort order for headers, as it's just a map with no guarantees about enumeration order.  (We could theoretically capture set order within the application for responses, but that's not necessarily transmittal order.) The proxies also do not, currently capture the exact capitalization of set headers (as the spec says they are case-insensitive way).

avanderhoorn:
Closing for the time being.

PR #144, opened 2016-06-27
Highlight request/response bodies based on content type

philliphoff:
A small change to the request/response Body sub-tab to highlight bodies based on the `Content-Type` header (e.g. use JSON syntax highlighting if the header indicates `application/json`).  Highlighting is done only for a set of well-known content types (easily added to as needed).

screen shot 2016-06-27 at 11 10 52 am

### DIFF
--- a/src/request/components/RequestDetailPanelRequest.tsx
+++ b/src/request/components/RequestDetailPanelRequest.tsx
@@ -2,6 +2,8 @@ import { TabbedPanel } from './TabbedPanel';
 import { TabPanel } from './TabPanel';
 import { trainCase } from '../../lib/StringUtilities';
 
+import requestConverter = require('../repository/converter/request-converter');
+
 import _ = require('lodash');
 import Highlight = require('react-highlight');
 import React = require('react');
@@ -49,7 +51,7 @@ export class Request extends React.Component<IRequestProps, {}> {
                         { this.renderHeaders(headers) }
                     </TabPanel>
                     <TabPanel header='Body'>
-                        { this.renderBody(body) }
+                        { this.renderBody(body, headers) }
                     </TabPanel>
                 </TabbedPanel>
             </div>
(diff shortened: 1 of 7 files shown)


philliphoff:
Associated with #131.

src/request/components/RequestDetailPanelRequest.tsx
```diff
@@ -72,11 +74,34 @@ export class Request extends React.Component<IRequestProps, {}> {
...
+        _.forEach(headers, (value, key) => {
+            if (key.toLowerCase() === 'content-type') {
+                contentType = value;
+                
+                return false;
+            }
+        });
+
+        return contentType;
+    }
```
avanderhoorn: Should the selector pick this out for us?

src/request/components/RequestDetailPanelRequest.tsx
```diff
@@ -72,11 +74,34 @@ export class Request extends React.Component<IRequestProps, {}> {
...
+
+    private getHighlightClassNameForContentType(contentType: string): string {
+        const category = requestConverter.getContentTypeCategory(contentType);
+
+        return (category && category.highlight) || '';
+    }
```
avanderhoorn: Here is an interesting question... given that the class name is very view related, I wonder if this should come out of the selector. My guess is not.

src/request/repository/converter/request-converter.js
```diff
@@ -39,16 +39,16 @@ var getContentTypeCategory = (function() {
...
+        'text/javascript':             {'script': true, 'highlight': 'javascript'},
+        'text/ecmascript':             {'script': true, 'highlight': 'javascript'},
+        'application/javascript':      {'script': true, 'highlight': 'javascript'},
+        'application/ecmascript':      {'script': true, 'highlight': 'javascript'},
+        'application/x-javascript':    {'script': true, 'highlight': 'javascript'},
+        'text/javascript1.1':          {'script': true, 'highlight': 'javascript'},
+        'text/javascript1.2':          {'script': true, 'highlight': 'javascript'},
+        'text/javascript1.3':          {'script': true, 'highlight': 'javascript'},
+        'text/jscript':                {'script': true, 'highlight': 'javascript'},
+        'text/livescript':             {'script': true, 'highlight': 'livescript'},
```
avanderhoorn: Good usage. I'm happy with this.

avanderhoorn:
Small comment about the placement of some code. But no major blocker.

src/request/components/RequestDetailPanelRequest.tsx
philliphoff: I'd agree--the use of Highlight.js is strictly an implementation detail of the view and so I'd want to limit exposure of aspects of that component (like the class name) to the view.  (Originally I had the whole mapping in this component as well, but then ran across a similar mapping in `request-converter` so just piggy-backed onto that instead.)

src/request/components/RequestDetailPanelRequest.tsx
philliphoff: I debated that, since the headers were already provided to the view and highlighting based on content type was also a kind of implementation detail of the view.  However, it's conceivable that we might want to determine the content type in other ways than the `Content-Type` header (e.g. its missing but we know it a priori or we sniff it instead), so it might be most useful to pass it to the view.  I've updated the PR.

PR #145, opened 2016-06-27
Support request/middleware headers schema change

philliphoff:
Corresponding change to [the Agent changes](https://github.com/Glimpse/Glimpse.Node.Prototype/pull/362) to enable the more accurate capturing of multiple header values.


src/request/components/RequestDetailPanelRequest.tsx
```diff
@@ -14,7 +14,7 @@ export interface IRequestProps {
     };
     response: {
         body: string;
-        headers: { [key: string]: string };
+        headers: { [key: string]: string | string[] };
```
avanderhoorn: Do we think its worth creating a type for `HeaderValues`? Might be easier than having to make sure that we repeat these semantics through the code.

avanderhoorn:
Good to go in when you are.

PR #147, opened 2016-06-29
Add visualization of overwritten/un-set headers

philliphoff:
Updates the Middleware table in the Request tab to visualize headers that were overwritten by following middleware or otherwise "un-set".  Also moves the Middleware UI to its own component.

screen shot 2016-06-29 at 11 15 06 am


philliphoff:
Related to #128.

src/request/components/RequestDetailPanelRequestMiddleware.tsx
```diff
@@ -0,0 +1,107 @@
...
+        );
+    }
+
+    private renderMiddlewareRow(ordinal: number, middleware: IFlattenedMiddleware) {
+        return (
+            <tr>
+                <td>{ordinal}</td>
+                <td>{this.renderName(middleware.middleware.name, middleware.depth)}</td>
+                <td>{middleware.middleware.packageName}</td>
+                <td>{_.size(middleware.middleware.headers) > 0 ? 'Header' : '-'}</td>
```
avanderhoorn: Thinking... this is something that would need to change if/when we track more than headers.

fake/fake-extension.js
```diff
@@ -217,9 +217,9 @@ var seedMvcActions = (function() {
...
-                            { access: 'middleware-end', correlationId: '4', name: 'jsonParser', displayName: 'JSON Body Parser', packageName: 'body-parser', paths: ['/'], result: 'next' },
+                            { access: 'middleware-end', correlationId: '4', name: 'jsonParser', displayName: 'JSON Body Parser', packageName: 'body-parser', paths: ['/'], headers: [{ op: 'set', name: 'content-type', value: 'application/json'}], result: 'next' },
                             { access: 'middleware-start', correlationId: '5', name: 'urlencodedParser', displayName: 'URL-Encoded Body Parser', packageName: 'body-parser', paths: ['/'] },
-                            { access: 'middleware-end', correlationId: '5', name: 'urlencodedParser', displayName: 'URL-Encoded Body Parser', packageName: 'body-parser', paths: ['/'], result: 'next' },
+                            { access: 'middleware-end', correlationId: '5', name: 'urlencodedParser', displayName: 'URL-Encoded Body Parser', packageName: 'body-parser', paths: ['/'], headers: [{ op: 'set', name: 'content-type', value: 'application/xml'}], result: 'next' },
```
avanderhoorn: I wonder if longer term we will put `header` as a property in `{ op: 'set', name: 'content-type', value: 'application/xml'}], result: 'next' }`, that way we could capture changes to `statusCode`, `body`, `cookies`, etc.

src/request/reducers/RequestDetailRequestReducer.ts
```diff
@@ -92,20 +75,37 @@ function correlateMiddlewareMessages(startMessages: IMessageEnvelope<IMiddleware
...
-        values, 
+        values,
         (result: { [key: string]: TResult }, value: T) => {
             result[keySelector(value)] = valueSelector(value);
 
             return result;
-        }, 
+        },
         <{ [key: string]: TResult }>{});
 }
```
avanderhoorn: I wonder how much having `toMap` abstracted out from `createMiddlewareHeaders` buys us?

src/request/selectors/RequestDetailRequestSelectors.ts
```diff
@@ -10,31 +9,50 @@ const getMiddlewareState = (state: IRequestState) => state.detail.request.middle
...
+        });
+
         middlewareArray.push({
             depth: depth,
-            middleware: {
-                name: middlewareItem.name,
-                packageName: middlewareItem.packageName,
-                headers: middlewareItem.headers
-            }
+            middleware: m
```
avanderhoorn: Maybe use `description` or something instead of `m`?

avanderhoorn:
Looks good. A couple of observations but nothing major.

src/request/components/RequestDetailPanelRequestMiddleware.tsx
philliphoff: Yes, the existing design also includes showing which middleware sets what cookies (not yet implemented).  I think the design is already a little awkward for those two concepts (headers and cookies); as we capture more (e.g. additions or changes to the request object), we'll need to rethink how they're all visualized in a single table form.

fake/fake-extension.js
philliphoff: I'd be a little concerned with the message schema becoming a little too "meta" and loose in that kind of an attempt to be able to describe every kind of change that might occur during middleware execution.  I think we'll still want to have concept specific sub-schemas for noting changes to specific objects (e.g. headers, status codes, body, cookies, etc.).

src/request/reducers/RequestDetailRequestReducer.ts
philliphoff: To an extent, it was just an exercise in seeing whether I could generally replicate the `ToDictionary()` extension method in .NET, as I was unable to find an equivalent in `lodash` (where you can take an array and create a map using independent selectors for key and value, all in a single step).  It mostly works, but the TypeScript compiler seems finicky about the selectors (e.g. it can't seem to infer the generic type of the method if you pass a predefined selector rather than an inline lambda, even if they're both typed the same way).

src/request/selectors/RequestDetailRequestSelectors.ts
philliphoff: Was originally `middleware` but the linter complained about it being a "shadow variable" so shortened it as I couldn't think of something better at the time.  I've updated it to `newMiddleware`.

src/request/components/RequestDetailPanelRequestMiddleware.tsx
avanderhoorn: Agreed.

fake/fake-extension.js
avanderhoorn: I can see that... At the same time we will want to keep an eye on ending up with a root level property for every type of change. Maybe we could have a root property call `changes` or `sideeffects` or something that could contain properties for things like headers, status codes, body, cookies, etc. All that said, totally no impact for this PR.

src/request/reducers/RequestDetailRequestReducer.ts
avanderhoorn: Wouldn't the equivalent be byKey? https://lodash.com/docs#keyBy

avanderhoorn: The one thing it doesn't do is provide a selector for the value.

philliphoff: > The one thing it doesn't do is provide a selector for the value.

Yep, and that's almost always what I need (and do, in this case). There are ways in `lodash` to pair and un-pair with a transform that work, but it's more processing and the TypeScript bindings make it less pleasant to work with.

Issue #148, opened 2016-06-29
Client does not work with Safari


philliphoff:
The existing client does not load when using Safari.  It looks like Safari is complaining about the use of a lambda in `request-view.jsx`.

(I may have broken this when inserting the Redux provider.)

PR #149, opened 2016-06-29
Replace lambda with anonymous function.

philliphoff:
Fix issue with the Client not able to load in Safari due to use of lambda function.  It appears that we aren't yet using Babel in the existing Client's Webpack build process and Safari v9 doesn't yet support lambdas (but v10 does).


philliphoff:
Associated with #148.

PR #150, opened 2016-06-30
Added support for parsing form data client-side

philliphoff:
While the `web-request` message defines properties corresponding to the form data parsed from a request body, and the current Glimpse Client can display such form data, the current Glimpse Agent does not currently perform such parsing.  This change updates the client to, if no form data exists but body content does exist and the content type is form data, parse the content itself.

screen shot 2016-06-30 at 8 46 38 am


philliphoff:
This is related to [this Glimpse Agent issue](https://github.com/Glimpse/Glimpse.Node.Prototype/issues/364).

src/request/reducers/RequestDetailRequestReducer.ts
```diff
@@ -150,7 +151,52 @@ const urlReducer = createRequestReducer<string>(
...
+
+    if (request && request.body && request.body.content && request.headers) {
+        const mediaType = getMediaType(getValueAtKeyCaseInsensitive(request.headers, 'content-type'));
+
+        if (mediaType.toLowerCase() === 'application/x-www-form-urlencoded') {
+            const formData = querystring.parse(request.body.content);
+
+            return formData;
+        }
+    }
```
avanderhoorn: Could these two if statements be merged a little cleaner?
```
    if (request && request.body) {
        if (request.body.form && !_.isEmpty(request.body.form) {
            return request.body.form;
        }
        if (request.body.content && request.headers) {
            const mediaType = getMediaType(getValueAtKeyCaseInsensitive(request.headers, 'content-type'));

            if (mediaType.toLowerCase() === 'application/x-www-form-urlencoded') {
                const formData = querystring.parse(request.body.content);

                return formData;
            }
        }
    }

```

avanderhoorn:
Looks good. Just one small comment.

src/request/reducers/RequestDetailRequestReducer.ts
philliphoff: Yep, I had intended to do exactly that last night (in the morning), but forgot.  Thanks for catching it.

avanderhoorn:
Assuming the tests pass, this is good to go.

PR #154, opened 2016-07-08
updating TelemetryClient to send more info.

mike-kaufman:
- Now sending serverID & other server properties retrieved from the TelemetryConfig resource.
 - Also fixed up logic around tracking the time spent on each tab and time spent on each request.


fake/fake-request-service.js
```diff
@@ -279,9 +285,9 @@ var details = (function () {
...
         }
     });
-    
+
     // stream subscribers
-    streamMock.on(window.location.origin + '/glimpse/message-stream/', function() {
+    streamMock.on(window.location.origin + '/glimpse/message-stream/', function () {
         summaries.stream();
     });
 })();
```
avanderhoorn: There are a fair few white space changes here. Are they all needed for this PR (specifically `function ()` vs `function()` - in most cases we use the later)?

fake/fake-request-service.js
```diff
@@ -234,33 +234,39 @@ var details = (function () {
...
                 "enabled": true,
                 "uri": "https://vortex.data.microsoft.com/collect/v1",
-                "instrumentationKey": "AIF-a96980ad-8a38-47a2-bbb0-328338b6964a"
+                "instrumentationKey": "AIF-a96980ad-8a38-47a2-bbb0-328338b6964a",
+                "clientIP": '0.0.0.0',
+                "serverMachineId": '0123456789abcdef',
+                "serverAppName": 'fakeapp',
+                "serverOSPlatform": 'fakeos',
+                "serverOSRelease": 'fakeosrelease',
+                "serverOSType": 'fakeostype'
```
avanderhoorn: I'm assuming this is waiting the PR from the server.

avanderhoorn: Don't worry, this fake data generator.

avanderhoorn:
Small question about what looks like some overly helpful auto formats but besides that GTG.

fake/fake-request-service.js
mike-kaufman: So, this was just the default formatting from vs code.  I can change it, but it raises the question of, if we're all using vs code & this is the default formatting rules, why are we using something different?

mike-kaufman:
Please take a quick look at latest updates.  I did the following:
 - updated build to generate a glimpseClientVersion file from what is in package.json
 - udpated client to send glimpse client version as part of the shell ready event. 
 - updated some property names for consistency, and normalized away some redundant properties.

glimpseClientVersion.js
```diff
@@ -0,0 +1,6 @@
+'use strict'
```
philliphoff: Should this be excluded from the repo (as it's intended to be generated)?

mike-kaufman: I thought about that, but thought that since it changes rarely, having it checked in would be provide more visibility to the file and when it changes.

gulpfile.js
```diff
@@ -129,13 +146,17 @@ gulp.task('clean', () => {
 });
 
 gulp.task('clean-test', () => {
-    return del(testSettings.output+ '/**');
+    return del(testSettings.output + '/**');
+});
+
+gulp.task('generateGlimpseClientVersion', (cb) => {
```
mike-kaufman: will rename.

glimpseClientVersion.js
philliphoff: The problem is that there's no guarantee that this file will get re-committed when regenerated.  For example, if I cut a release, I first update the version number in `package.json`.  I don't do a build myself, because when that PR is merged I'll then create a release in GitHub.  The CI system will do the actual build that generates the NPM package (indirectly updating that file), but the CI system won't submit that change back to the repo.

mike-kaufman:
OK, removed glimpseClientVersion.js from repo & added it to .gitignore.

PR #157, opened 2016-07-12
bumping client version to 8a

mike-kaufman:
Wanted to get as many people as possible using the new client which has updated telemetry events.  This way we can verify the backend pipeline is working correctly.

### DIFF
--- a/src/lib/TelemetryClient.ts
+++ b/src/lib/TelemetryClient.ts
@@ -349,7 +349,22 @@ class TelemetryClient {
      * then the specified props will be updated.  Otherwise, a new instance will returned. 
      */
     private getShellReadyProperties(props?: IShellReadyProperties): IShellReadyProperties {
-        props = props || <IShellReadyProperties>{};
+
+        if (!props) {
+            const p: IShellReadyProperties = {
+                sessionId: undefined,
+                clientCookieID: undefined,
+                glimpseClientVersion: undefined,
+                glimpseServerVersion: undefined,
+                serverMachineId: undefined,
+                serverAppName: undefined,
+                serverOSPlatform: undefined,
+                serverOSRelease: undefined,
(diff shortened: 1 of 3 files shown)


mike-kaufman:
Also through in a fix for a missing property in one of the telemetry events.  This wasn't getting caught by typescript build due to a cast, so I fixed that so build should catch further issues like this.

src/lib/TelemetryClient.ts
```diff
@@ -349,7 +349,22 @@ class TelemetryClient {
      * then the specified props will be updated.  Otherwise, a new instance will returned.
```
philliphoff: Just curious as to the reason for overwriting passed-in values (as opposed to the general JavaScript convention that passed in values are not overwritten).

philliphoff:
Just one minor question, but otherwise LGTM.

src/lib/TelemetryClient.ts
mike-kaufman: The "telemetryConfig" instance is retrieved asynchronously, often after the "shell ready" event is queued up, so we need to backfill a lot of values here.  I wanted one method to either create this instance, or backfill an existing instance.

If you're question is why are we updating the passed in props, as opposed to doing a shallow copy, no good reason other than it is slightly more efficient.  Given the narrow (private) uses of this function, I don't think this is subject to misuse or confusion, but I can change if you feel strongly about it.

philliphoff: As long as the expected usage is clear (and not widespread), it's probably fine.  It just struck me as a "JavaScript usually doesn't work this way" kind of a thing and so I had to spend extra time to work out the intent behind it.

mike-kaufman:
OK, I'll leave as-is for now, and try to fix up when I'm in here again.

